PR #175, opened 2014-11-22
Unity Project

somdoron:
Because of the demand I added a unity project, in next nuget release a unity package will be included.

Ready to be merge


krisrok:
what is delaying the merge? the added unity projects seem not to affect the other projects.

somdoron:
I think it is actually not needed, the v3.3.0.12 should work on unity with
no iisues

krisrok:
seems like it :)

PR #206, opened 2015-01-20
Making sure that NetMQMonitor can always be stopped/disposed

ndextraze-pbp:
This fix will ensure that the NetMQMonitor can be Stopped/Disposed even if an exception is raised from the MonitoringSocket.

It would interesting to know why Disconnect fails in this specific case. I don't know the internals of NetMQ enough. Maybe it's the designed behavior. In any case the NetMQMonitor should always complete properly (setting IsRunning = false and isStoppedEvent).

### DIFF
--- a/src/NetMQ.Tests/MonitorPollTests.cs
+++ b/src/NetMQ.Tests/MonitorPollTests.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
+using System.Security;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


somdoron:
thanks

somdoron:
@ndextraze-pbp the test is not stable, it failed from time to time. Can you please check? http://teamcity.codebetter.com/viewLog.html?buildId=179399&buildTypeId=bt1046&tab=buildResultsDiv

ndextraze-pbp:
I'll fix it, I tested too much. Like I said I don't have a good understanding of netmq the internals, but I'm starting to learn.

So the original issue happened if I understand correctly because the monitored socket got reaped before the monitor socket gets disposed. So the pipe between them gets destroyed by the monitored socket reaping and then the monitor disposal try to unregister that same pipe on its side but it's already gone. So the try finally fixed that.

I was letting the exception go free with the try-finally and expecting an EndpointNotFoundException to fault the monitor thread. The issue is that the reaping of the monitored socket is done async by the reaper so it is possible that the monitored socket get reaped after the monitor socket dispose. That's why we don't always see that exception.

I guess the best approach would be to try catch and just swallow any exception, since at this point it doesn't matter much.

Will provide a pull request soon.

ndextraze-pbp:
Pull request https://github.com/zeromq/netmq/pull/214

somdoron:
@ndextraze-pbp thanks. Do you want to join NetMQ maintainers? it is mainly reviewing and merging pull requests. we are following http://rfc.zeromq.org/spec:22.

ndextraze-pbp:
I have read the C4 process and I accept.

somdoron:
great. I added you to the maintainers list. You can start by review the current pull requests.

PR #207, opened 2015-01-21
release buffer after processing

somdoron:


somdoron:
ready to be merged

somdoron:
@reiroldan or @krisrok and this as well

ndextraze-pbp:
I'm getting PollTests failures after on master TC build after merging this pull request. Should I revert?

somdoron:
It is a known isssue, the two timers test fail from time to time because it needs to be accurate. I need to fix the test somehow

PR #216, opened 2015-01-25
simplfy two timers test

somdoron:
This test failed from time to time because it depends on how accurate the code is being cold. I simply the test so it won't be depended on accuracy.


somdoron:
@reiroldan, @krisrok or @ndextraze-pbp can you please merge?

PR #218, opened 2015-01-30
Update form base

BruceMellows:


somdoron:
@BruceMellows I cannot merge this pull request. it break the entire package and is not backwards compatible.

I if you want to a way to make netmq more testable I suggest just making an interface for the socket (and not changing the create methods).

Regarding factory instead of Context, you can create one, or create it in your own project. But don't change the tests for existing classes. it will break existing projects.

BruceMellows:
Sorry, I didn't intend for you to merge that. I was actually trying to
merge the base changes into my fork and slipped and hit my git newby head.

PR #235, opened 2015-02-19
Fix typo. Receivev* -> Receive*

drewnoakes:
This changes the public API, but in quite an obvious way. Instead the old symbols could also be retained and marked obsolete. Let me know if that's preferred and I'll update the PR.


somdoron:
I think it better to not break the old code, mark existing as obsolete and add new properties

drewnoakes:
Sure thing. Will fix in the next day or two.

drewnoakes:
Ok, I think this is good to go now.

PR #243, opened 2015-02-21
Add URL to LGPLv3 license in NuGet spec

drewnoakes:


somdoron:
because the netmq license has an exception I think it better to link to our license at:

https://github.com/zeromq/netmq/blob/master/COPYING.LESSER

drewnoakes:
Ah, I didn't know there was an exceptional clause. Commit amended and pushed.

PR #248, opened 2015-02-21
Annotations, minor tidying

drewnoakes:
Small things spotted while reading through the code.

This should be the last odd-bits PR.


drewnoakes:
As far as I can tell, the test failure is due to timing expectations combined with non-deterministic execution. I don't think I can manually kick off another build on Team City.

drewnoakes:
Force-pushed a minor change in order to trigger another build.

PR #259, opened 2015-02-26
XML-comments part 1 by James Hurst

JamesWHurst:
I've gone through about half the code - adding comments at those places where I could readily ascertain exactly what that property or method was for.  The intent is to enable the Visual Studio Intellisense to help guide the developer in his coding when she makes use of this API. The clear wording of the API itself helps to a point, of course - but I've tried to put comments that added clarity, hints, specified what the default values are of properties, etc. 
I've also commented every Exception subclass, and added a concise Message at just about every point at which an exception is explicitly thrown (for that code that I did get through).

I skipped those places where I just don't know exactly what the code is doing or why. Better to leave that for later, or for someone else. No comment is better than one that misses an important point.

This takes massive time. So I'd like to get feedback, before I do more of it.

### DIFF
--- a/src/NetMQ.Tests/ActorTests.cs
+++ b/src/NetMQ.Tests/ActorTests.cs
@@ -1,9 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using NetMQ.Sockets;
-using NUnit.Framework;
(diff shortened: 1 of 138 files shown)


somdoron:
@JamesWHurst amazing job!!

So I went over most of the comments, seems great and correct.

PR #262, opened 2015-02-27
Added XML-comments, part 2, added Utils/StringLib

JamesWHurst:
Further refinement of the XML (Intellisense-compatible) comments. I also added a file, zmq/Utils/StringLib.cs, to contain string-related misc utility methods.


somdoron:
amazing, great contribution. Thanks @JamesWHurst

PR #270, opened 2015-02-28
Fixed: IDisposables not being disposed.

tobi-tobsen:
Some IDisposables were not properly disposed.


somdoron:
@tobi-tobsen the changes break some tests, can you check it out?

tobi-tobsen:
Yes, of course! Somehow these were not breaking on my machine. Will try to reproduce and fix.

tobi-tobsen:
I was able to reproduce when compiling "mixed platforms". Strangely it compiled and tested fine with x86 yesterday. Will have a look into it later.

PR #273, opened 2015-03-01
Merged latest changes with comments I added, and resolved a few compile-warnings.

JamesWHurst:
To cut down on the work of merging in my pull-request, I merged in the latest changes as of today, March 1st (rather substantial!).
Also updated the help-file to reflect the latest code (that gets generated by automatically sucking-in the XML-formatted comments from the source-code using SandCastle).


drewnoakes:
Wow you've really been busy with documenting the code! Great stuff.

As for the generated CHM file, I think we should not include in the repository. Git doesn't deal well with large binary files -- it is far better suited to easily diff-able text files.

The CHM file merged in already is 1.5MB. If we include an updated help file every so often, the repo size will grow very quickly such that cloning it becomes very slow.

I'd suggest we include instructions on how to generate this CHM file instead.

If you agree, could you rework this PR to exclude the first commit? You can do this relatively easily with interactive rebase.

I'm also not a huge fan of grouping code into regions of public/private or properties/methods other than when having them ungrouped hides the pattern of the class somehow. The existing code does not really make much use of regions. My recent commits relating to code style have been trying to move the code towards a single consistent style based on the style used predominantly throughout, even though that is not my own first choice. The style should at least be internally consistent.

krisrok:
wanted to chime in thanking you for your effort, james. 

also, i agree we should not add the chm to the repo. ways to generate it may be a nice to have feature, but i guess it would go mostly unnoticed as it means additional user effort. you should be easily able to come by help.

afaik chm is only supported natively by windows anyway. maybe we should choose a different format, like some offline html docs (what are the cool kids using these days?).

or is there anything special the chm format has to offer? else from already being implemented by james :/

JamesWHurst:
I agree - it really doesn't make sense to have a generated help-file be version-controlled. If we had a directory for releases, that I think would be the logical place. It's not source, it is end-product, but it's useful for end-users to have without having to generate it for themselves.
And it *is* Windows-only, so it should go into a Windows-only release directory. A cross-platform format such as plain HTML would be better. I'm not sure how to produce that just yet. This .chm file is the original Windows Help 1 format, viewable on all versions of Windows going back to XP.

drewnoakes:
The simplest thing is to ship the XML docs along with the binaries in the NuGet package.

drewnoakes:
Just looking into merging this. I can pick the one commit (8462ed2), but I think it undoes some stuff, losing changes when you merged. For example, in Actor.cs. It's also a little hard to see what's going on when members were reordered. And there's quite a bit of indentation broken in v0_1/HandshakeLayer.cs, for example.

I noticed a few places in your commits where `<see cref="Foo"/>` was replaced with `Foo`. Is there a reason for this? Perhaps something to do with CHM compilation? If not, then it's good to use the cref attribute so that refactorings (such as rename) keep the documentation up to date and correct.

I am working on picking out the intended changes now and will push them separately if all goes well.

drewnoakes:
Ok, I fixed the conflicts. It's pushed as b4b76897e.

Thanks again.

JamesWHurst:
Thanks y'all.  Q regarding the help-file..  I'm not versant with the mechanics of creating NuGet packages (yet). How would one go about adding a help-file to that? Or is that even appropriate - since the NuGet packages would (I would think) be auto-updating by build servers and the like, who have no need of a help-file. ?
If it's too much trouble to include it as a download-link on on GitHub page itself, then perhaps I can just host the help-file on my own website and you could provide a link to it on the GitHub project home-page?

In the meantime, I'm going to sniff around for a tool that can create an HTML help-file equivalent, so that I can just host that and solve the portability problem. 

The one advantage of the stand-alone helpfile, is that you can reference it even when your connection is down, and it's a pretty easy and standard format that developers are used to from other languages. It's what I was wishing for when I first starting exploring NetMQ. Plus it shows up pretty quickly which global things are still without comments.

PR #274, opened 2015-03-01
Fix #272. More unit tests. Some minor tidying.

### LINKED ISSUE #272: Lazy receive sequence is not re-entrant
The lazy receive functions described in #64 and implemented in #65 have a potentially surprising pitfall.

Because they yield a sequence of results directly "from the wire", the sequence can only be enumerated once. This is, generally, not true of `IEnumerable<T>` types throughout the BCL.

Here is our `ReceiveMessages` method:

    public static IEnumerable<byte[]> ReceiveMessages(this IReceivingSocket socket)
    {
        bool hasMore = true;

        while (hasMore)
            yield return socket.Receive(false, out hasMore);
    }

However user code like this doesn't actually work:

    var frames = sock.ReceiveMessages();
    if (frames.Any())
    {
        var header = frames.First(); // WRONG
    }

In this case, `Any` burns the first item in the sequence, and `First` actually gives the _second_ frame (or `throw`s if there was only one frame in the message).

What benefit is there of dealing with a sequence? As ZeroMQ only deals in full messages, aren't frames already buffered until a whole message is available on a given socket?

Generally I'm a fan of making things lazy, however when the consequence of multiple enumeration is not simply a performance hit but in fact completely incorrect results, I think laziness is not a good fit.

I propose modifying `ReceivingSocketExtensions` to always return materialised collections, and do so using `IList<byte[]>`.

Thoughts?

drewnoakes:
Please review especially:

* 58cab65 fix for #272
* f6593a8 removal of destructors

### DIFF
--- a/src/NetMQ.Tests/ActorTests.cs
+++ b/src/NetMQ.Tests/ActorTests.cs
@@ -10,7 +10,7 @@ namespace NetMQ.Tests
         {
             using (var context = NetMQContext.Create())
             {
-                using (NetMQActor actor = NetMQActor.Create(context, shim =>
+                ShimAction shimAction = shim =>
                 {
                     shim.SignalOK();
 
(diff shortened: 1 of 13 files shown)


JamesWHurst:
Very good improvements - glad you're attending to these rough edges.

PR #281, opened 2015-03-03
Fix recently broken null-test and use correct annotation.

drewnoakes:
My bad. I took the annotation at face value. Discovered today that the value can be null (when writing identity), so fixed this up.


JamesWHurst:
I appreciate your fastidious attention to these details  (been wanting to use that word all day).

PR #286, opened 2015-03-06
Workbranch

JamesWHurst:
This pull-request is to pull from the "workbranch" branch of my repo.
This is mostly corrections of spelling errors, such as renaming the Initialised to Initialized, and the comments I've been adding.
I merged this with the current state of the src-code as of today, to minimize the work needed to resolve this pull-request.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -4,7 +4,7 @@ Contributors
 * Alexey Yakovlev &mdash; [@yallie](https://github.com/yallie)
 * Christophe Juniet &mdash; [@cjuniet](https://github.com/cjuniet)
 * Dan de Zille &mdash; [@dandezille](https://github.com/dandezille)
-* Dorom Somech &mdash; [@somdoron](https://github.com/somdoron)
+* Doron Somech &mdash; [@somdoron](https://github.com/somdoron)
 * Drew Noakes &mdash; [@drewnoakes](https://github.com/drewnoakes)
 * Eugen Dück &mdash; [@EugenDueck](https://github.com/EugenDueck)
 * Grant Birchmeier &mdash; [@gbirchmeier](https://github.com/gbirchmeier)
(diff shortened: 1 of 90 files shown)


drewnoakes:
Apologies for the delay in looking at this. It's been a busy week.

I added a few comments inline. If you'd like, I can go through and pull out the parts of this which don't break the API and rebase them on top of current master.

As a speaker of British English I am not so willing to agree that _initialised_ is a mistake :)

JamesWHurst:
Agreed. I put it threw a spelling-checker (VSSpellChecker) which obviously needs a better perspective on what is and is not correct. Yes - can you merge this in without the bits that you mention above?

JamesWHurst:
While on the topic of changes - why do we have two Poller classes?  Yes, one is an internal class - so there's no public conflict. But can we rename it to something that distinguishes it in purpose from the public Poller class, to avoid confusion? And.. how would you describe it's purpose and operation, in a nutshell - as opposed to the public Poller?  I don't want to add brain-dead comments, but could do with some clarification.  I am assuming that renaming an internal class does not break our API. ?

JamesWHurst:
I appreciate your enlightening me. Found this reference:  http://english.stackexchange.com/questions/60852/queueing-or-queuing
  -which btw has a delightful page-design and typeface.  You're totally right - these are correct spellings, and I do like the English version better
Hold off a bit -- I'll go ahead and fix them and re-commit.  Q: when you do a commit to your fork, AFTER initiating a pull-request, does have the same result as though you had done the commit *before* you initiated the pull-request?

drewnoakes:
> Why do we have two Poller classes?

The `NetMQ.zmq` namespace contains duplicates of many public types. It seems to be a legacy of the code's origins. I've wondered whether the parallel hierarchies may be merged at some point, but it'll be non-trivial to get right, and also require making the `ZObject`/`Own`/etc classes public, or hiding them internally via composition (a la pimpl-idiom). The parallel socket types are another good example.

I'm not sure how to document them, and as they may go away one day it may not be worth documenting the internal duplicates. Focussing our energies on comprehensive documentation of the public API would be better for the users.

I'd like to produce a developer's guide to the internals of the library at some point, though this will require my gaining significantly greater understanding of these internals.

> When you do a commit to your fork, AFTER initiating a pull-request, does have the same result as though you had done the commit before you initiated the pull-request?

I didn't get to answer this in time, but I see you found the answer all the same :)

PR #289, opened 2015-03-09
Fix code analysis paths

sean-gilliam:
This fixes error CA0053 where visual studio was unable to locate the
static analysis tools used by Code Analysis.


drewnoakes:
Thanks.

Issue #275, opened 2015-03-01
Messages, Frames and the socket receiving APIs


drewnoakes:
The extension methods for receiving from sockets are not very friendly when starting out with NetMQ. There are several ways to achieve the same thing (`SendReiveOptions.DontWait` vs `bool dontWait`) and the behaviour of non-blocking reads are not clear (do they return `null`, throw, or something else?)

In the ZeroMQ Guide, there's a distinction between messages and frames. NetMQ's API commonly refers to frames as messages. ~~This is pretty confusing and it'd be nice to get away from.~~ _Turns out that multipart message support is slated for removal in the future of ZeroMQ._

Here are some ideas for how the receiving APIs might work in future.

```csharp
//////// BLOCKING APIS

// Receiving frames

Frame frame; // rename of the current Msg class
socket.ReceiveFrame(ref frame); // void return

byte[] frame = socket.ReceiveFrameBytes();
byte[] frame = socket.ReceiveFrameBytes(out more);

string frame = socket.ReceiveFrameString();
string frame = socket.ReceiveFrameString(out more);
string frame = socket.ReceiveFrameString(Encoding);
string frame = socket.ReceiveFrameString(Encoding, out more);

// Receiving entire messages

NetMQMessage message = socket.ReceiveMessage();

List<byte[]> message = socket.ReceiveMessageBytes();

List<string> message = socket.ReceiveMessageStrings();
List<string> message = socket.ReceiveMessageStrings(Encoding);

//////// NON-BLOCKING APIS

// Receiving frames

Frame frame; // rename of the current Msg class
bool success = socket.TryReceiveFrame(ref frame); // void return

byte[] frame;
bool success = socket.TryReceiveFrameBytes(out frame);
bool success = socket.TryReceiveFrameBytes(out frame, out more);
bool success = socket.TryReceiveFrameBytes(TimeSpan, out frame);
bool success = socket.TryReceiveFrameBytes(TimeSpan, out frame, out more);

string frame;
bool success = socket.TryReceiveFrameString(out frame);
bool success = socket.TryReceiveFrameString(out frame, out more);
bool success = socket.TryReceiveFrameString(out frame, Encoding);
bool success = socket.TryReceiveFrameString(out frame, Encoding, out more);
bool success = socket.TryReceiveFrameString(TimeSpan, out frame);
bool success = socket.TryReceiveFrameString(TimeSpan, out frame, out more);
[...]

PR #292, opened 2015-03-11
Tweaks

drewnoakes:
A bunch of small improvements made during work on #275. The pending receive/send API work for that issue depends upon these tweaks, and I'd like to make that a lean PR, hence pushing these little bits up separately.

Particularly worth reviewing:

* cfcfa3c48af0700486c94e24351ed85b0a11dcb2
* bbd0e940fe307902e1cefbf5f80d45c67026657f
* cbf92e7fe1196596afd6268919eb795dc8a93615
* 6b976c385748cfb2c13f3f21e84ce3e34b43e3ea

It's probably easiest to review each commit in isolation, as each has a distinct theme.

### DIFF
--- a/src/NetMQ/ReceivingSocketExtensions.cs
+++ b/src/NetMQ/ReceivingSocketExtensions.cs
@@ -66,14 +66,7 @@ namespace NetMQ
         [NotNull]
         public static byte[] Receive([NotNull] this IReceivingSocket socket, bool dontWait, out bool hasMore)
         {
-            var options = SendReceiveOptions.None;
-
-            if (dontWait)
-            {
-                options |= SendReceiveOptions.DontWait;
-            }
-
-            return socket.Receive(options, out hasMore);
+            return socket.Receive(dontWait ? SendReceiveOptions.DontWait : SendReceiveOptions.None, out hasMore);
         }
 
         /// <summary>
@@ -85,7 +78,7 @@ namespace NetMQ
         [NotNull]
         public static byte[] Receive([NotNull] this IReceivingSocket socket, out bool hasMore)
         {
-            return socket.Receive(false, out hasMore);
+            return socket.Receive(SendReceiveOptions.None, out hasMore);
         }
 
         /// <summary>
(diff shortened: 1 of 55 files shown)


drewnoakes:
Whoops. Looks like I didn't separate the changes cleanly. Will push new commits. Don't merge this yet!

drewnoakes:
Ok that should be better. Never push then go to bed, or you may wake up to a surprise :)

src/NetMQ/ReceivingSocketExtensions.cs
```diff
@@ -393,13 +370,9 @@ public static void ReceiveMessage([NotNull] this IReceivingSocket socket, [NotNu
...
 
-            bool more = true;
-
+            var more = true;
             while (more)
-            {
-                byte[] buffer = socket.Receive(dontWait, out more);
-                message.Append(buffer);
-            }
+                message.Append(socket.Receive(dontWait, out more));
```
somdoron: here we can call the method with the Msg parameter and improve performance a little by reusing the msg structure.

somdoron:
@drewnoakes can I merge?

src/NetMQ/ReceivingSocketExtensions.cs
drewnoakes: Good spot. I've a complete overhaul of these methods coming in a subsequent
PR and will make sure this is included.

[via phone]

drewnoakes: Yes I believe it's OK, if you agree.

(Messages out of sync as I was underground with no signal)

[via phone]

drewnoakes:
Yes I believe it's OK, if you agree.

src/NetMQ/ReceivingSocketExtensions.cs
somdoron: I also think we should make the NetMQMessage disposable (including frames) and then use the Msg inside the NetMQFrame, it will save copy and allocation. But this will change the behavior of NetMQMessage a little, so we need to think how to do that (maybe new class). For example now after sending the message the message will be cleaned (we don't own the msg after sending) which is different behavior.

I believe we can make Message class with good performance (minimal allocation and copying) which be very good alternative to send/recv methods.

PR #296, opened 2015-03-13
New receive APIs

drewnoakes:
As discussed in #275, this PR adds new methods for receiving data from `IReceivingSocket` and marks all old receive methods as obsolete.

This change moves away from using `AgainException` to signal timeouts when receiving.

Receive methods have two flavours. For example:

```csharp
var data = socket.ReceiveFrameString();       // blocks indefinitely
    
string data;
if (socket.TryReceiveFrameString(out data))   // non-blocking
    Console.WriteLine("Received " + data);
else
    Console.WriteLine("No message available");
```

This necessitated marking `SocketOptions.ReceiveTimeout` obsolete. The blocking APIs cannot be interrupted without using exceptions, and in general exceptions are slow and harder to reason about. Users of the socket-level receive timeout should pass timeout durations to the `TryReceive` methods:

```csharp
socket.TryReceiveFrameString(TimeSpan.FromSeconds(10), out data)
```

Please try out these APIs and provide feedback. They may continue to evolve freely before the next release, after which they will become more difficult to change without adding even more `[Obsolete]` attributes.


src/NetMQ/IReceivingSocket.cs
```diff
@@ -1,18 +1,62 @@
...
+        
+        /// <summary>
+        /// Block until the next message arrives, then make the message's data available via <paramref name="msg"/>.
+        /// </summary>
+        /// <remarks>
+        /// The call  blocks until the next message arrives, and cannot be interrupted. This a convenient and safe when
+        /// you know a message is available, such as for code within a <see cref="NetMQSocket.ReceiveReady"/> callback.
+        /// </remarks>
+        /// <param name="msg">An object to receive the message's data into.</param>
+        void Receive(ref Msg msg);
```
somdoron: Why is this part of the interface, isn't it just a call to TryReceive with Infinite as timeout? it can be an extension method and simplify the API

src/NetMQ/ReceivingSocketExtensions.cs
```diff
@@ -8,326 +9,999 @@
...
+        /// </summary>
+        /// <param name="socket">The socket to receive from.</param>
+        /// <param name="encoding">The encoding used to convert the frame's data to a string.</param>
+        /// <param name="expectedFrameCount">Specifies the initial capacity of the <see cref="List{T}"/> used
+        /// to buffer results. If the number of frames is known, set it here. If more frames arrive than expected,
+        /// an extra allocation will occur, but the result will still be correct.</param>
+        [NotNull]
+        public static List<string> ReceiveMultipartStrings([NotNull] this IReceivingSocket socket, [NotNull] Encoding encoding, int expectedFrameCount = 4)
+        {
+            var frames = new List<string>(expectedFrameCount);
```
somdoron: we can actually improve the performance by repeating a code but reusing the Msg

src/NetMQ/ReceivingSocketExtensions.cs
```diff
@@ -437,24 +1358,124 @@ public static bool WaitForSignal(this IReceivingSocket socket)
...
+            if (!socket.TryReceive(ref msg, timeout))
+            {
+                more = false;
+                msg.Close();
+                return false;
+            }
+
+            more = msg.HasMore;
+            msg.Close();
+            return true;
```
somdoron: another idea if you did skip frame is to do DropAll which drop the rest of the frames (until more). It is actually very useful because when you implement a protocol and there is an error you want to drop the entire frames in the message and continue to next message

somdoron:
looks good, do you want me to merge or to check the comments first?

drewnoakes:
I'll be out for a few hours but will integrate your feedback when home
again and extend this PR, so let's hold off on merging for now.

[via phone]

src/NetMQ/IReceivingSocket.cs
drewnoakes: I like this proposal. It is simpler. I was thinking that there may be some future optimisation for the indefinite blocking case, but in practice it could just a single branch per receive call.

src/NetMQ/ReceivingSocketExtensions.cs
drewnoakes: Nice. a3bae1f14cd460.

src/NetMQ/IReceivingSocket.cs
drewnoakes: eecc792cc9b33

src/NetMQ/ReceivingSocketExtensions.cs
drewnoakes: 20388271ea2e4

drewnoakes:
> We can actually do that with a trick, when in debug we can create a class and save the reference to it, the class will hold the status of the msg (and will be updated on changes), when ever the class is collected it have to be not initialized.

That's an elegant solution. I'll look at that now. Unfortunately we won't have stack traces for instances (though we could query them, maybe if a new compiler symbol is set) but at least we'd know there were unclosed messages.

Another option may be to create an implementation of `IBufferPool` that keeps track of where pooled buffers were not released. Is correct pool usage the only reason for being so careful with the `Msg` lifecycle?

somdoron:
In general yes, but I think it is important also for the health of the
library and future development

somdoron:
let me know when the PR is ready for merge

drewnoakes:
Your feedback has now been integrated, and the conflicts addressed. I'm happy with this PR now if you are.

somdoron:
great

JamesWHurst:
If y'all want to consider it -- I'd vote for leaving out the Encoding parameter. That only multiplies the number of methods, and the number is already considerable. Generally, I think a developer makes her choice of Encoding at the beginning of a project, and then that never changes. I cannot think of any reason it should ever change. Just a simple, single static property, or perhaps an instance-property on the context -- surely not mixed-in with the receive methods.  This is already new (and better) API, why not keep it as clean and simple as possible?
I have never, ever known a case where code would need to switch among different encodings during the course of a transaction.  Especially if the default is UTF-x

drewnoakes:
@JamesWHurst, I agree that there are a lot of overloads and it'd be nice to reduce the count, but string encoding is something that we as native ASCII speakers don't come up against that often :) Hopefully the overloads remind developers that byte-to-string conversion is not always 1:1. Also it is possible, however unlikely, that a consumer must use different encodings for different frames of the same message sent by a non-NetMQ peer.

drewnoakes:
@JamesWHurst I came across [this interesting article](https://modelviewculture.com/pieces/i-can-text-you-a-pile-of-poo-but-i-cant-write-my-name) yesterday about character encodings. It shows why UTF-8 is not necessarily the last word in character encodings.

JamesWHurst:
That's a very interesting and illuminating article - thanks. Also disappointing.  I will be very keen to see Unicode evolve to be a better solution: I love languages and script and the exploration of them, so this article hits home to me.
I do believe, however, that as of today and the forseeable future, we are not likely to see a more suitable substitute. Unicode (UTF-8, UTF-16 or UTF-32 - regardless) is the standard closest to universal adoption at this point, and it clearly is far superior to our ASCII encoding. The only drawback that I can imagine, is the fact that we can no longer bank on the strings mapping 1-to-1 with bytes. And that's another reason to change the default, and to do it as soon as possible: that would ensure that all future development, debugging and fine-tuning, has to take that into account.  My own languages that I'm struggling to learn, include Russian, Mandarin, and Arabic (MSA). The software that I am designing, has to enable conversations in all of these - simultaneously. It'd be impossible with Unicode.

PR #317, opened 2015-03-31
Fixed invalid TimeSpan values which cause infinite timeout for polling t...

AlekseiOhotnik:
Fixed invalid TimeSpan values which cause infinite timeout for polling to become 0 or not -1 as expected with high cpu and memory pressure as side effect.
In one case FromTicks(-1) was used which didn't work properly as code down the callstack was using timeout.TotalMilliseconds with result of 0. In other case TimeSpan.MinValue was used with similar outcome but on Socket.Select or Socket.Poll methods. This problem is not going to be evident unless monitoring cpu usage or memory allocations, as no noticable functional issues seem to occur.
Issue was noticed on a single socket service with cpu spikes and 10mb/s memory allocs after updating to sought after implementation of TryRecieve methods.

### DIFF
--- a/src/NetMQ/Core/SocketBase.cs
+++ b/src/NetMQ/Core/SocketBase.cs
@@ -887,14 +887,14 @@ namespace NetMQ.Core
         /// </summary>
         /// <param name="msg">the <c>Msg</c> to read the received message into</param>
         /// <remarks>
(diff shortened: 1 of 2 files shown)


drewnoakes:
Cheers. Looks like some unit tests are needed for these functions.

PR #318, opened 2015-03-31
Changed lists to use Count instead of Any

AlekseiOhotnik:
While doing profiling usage of Any() surfaced as memory footprint in #Gen0. Since the support of explicit Count property is there it should be used instead.


drewnoakes:
Looks good. Thanks.

PR #320, opened 2015-04-02
Found that when running pub/sub sockets on a Raspberry PI subscription requests were never processed.

wrdouglass:
Found that when running pub/sub sockets on a Raspberry PI subscription requests were never processed.

The root cause was that RdtscOnArm was returning tick counts (milliseconds) and the throttling code was expecting high-resolution CPU timings. This caused the socket to be throttled virtually indefinitely (MaxCommandDelay = 3000000 milliseconds).

Simplified the ARM detection code to instead just set s_rdtscSupported = false which disables the throttle checking for ARM processors and removed associated code that was no longer used.

### DIFF
--- a/src/NetMQ/Core/Utils/Clock.cs
+++ b/src/NetMQ/Core/Utils/Clock.cs
@@ -23,90 +23,89 @@ using System.Diagnostics;
 
 namespace NetMQ.Core.Utils
 {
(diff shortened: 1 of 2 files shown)


wrdouglass:
I botched the tab v space requirement. fixing.

PR #329, opened 2015-04-13
Titanic Protocol Example

mph911:
created an example using Majordomo Protocol -> Titanic Protocol

PR #330, opened 2015-04-14
Titanic Protocol utilizing Majordomo Protocol -> Example

mph911:
see #329 but removed packages/* according to Drew Noakes request.

Sorry for bloating!

Peter


drewnoakes:
This looks good, but I'm going to squash the commits to ensure the packages folder doesn't get into the history (and increase the size when cloning the repo). Hope to do so this evening.

drewnoakes:
Merged into master now. Thanks very much!

PR #332, opened 2015-04-16
Use JetBrains Annotations from nuget

lufka:


drewnoakes:
This is definitely a better way to get these annotations into the project. Note that this does not create a runtime dependency on the `JetBrains.Annotations.dll` as all attributes are `[Conditional(...)]`, so it shouldn't affect library users.

I'm not sure why the build is failing here. It works on my PC, where VS goes off and fetches the NuGet dependency before building.

@somdoron, did you have to do anything special to the build to have it resolve the `AsyncIO` package?

leverate-somech:
log is full of this: InProcActors\Actor.cs(3, 7): error CS0246: The type or namespace name 'JetBrains' could not be found (are you missing a using directive or an assembly reference?)

leverate-somech:
ok because the dll is not in the packages we now need to change the build to also do nuget restore

drewnoakes:
@somdoron, yep. VS does this restore automatically. Maybe the build method is different on Team City.

src/NetMQ/NetMQ.csproj
```diff
@@ -44,6 +44,9 @@
       <HintPath>..\packages\AsyncIO.0.1.17.0\lib\net40\AsyncIO.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="JetBrains.Annotations">
+      <HintPath>..\..\..\..\packages\JetBrains.Annotations.9.1.1\lib\net20\JetBrains.Annotations.dll</HintPath>
+    </Reference>
```
leverate-somech: I think it should be ..\packages\JetBrains.Annotations.9.1.1\lib\net20\JetBrains.Annotations.dll

leverate-somech:
@lufka I think I found the issue, left you a comment in the log regarding the path of the annotations reference.

chinattw:
NETMQ Running!
WIN7 IS Succeed
Windows XP SP3 IS Faild

ERROR INFO
Framework Version: v4.0.30319
Exception Info: System.EntryPointNotFoundException
Stack:
   at AsyncIO.Windows.UnsafeMethods.GetQueuedCompletionStatusEx(IntPtr, IntPtr, Int32, Int32 ByRef, Int32, Boolean)
   at AsyncIO.Windows.CompletionPort.GetMultipleQueuedCompletionStatus(Int32, AsyncIO.CompletionStatus[], Int32 ByRef)
   at System.Threading.ThreadHelper.ThreadStart_Context(System.Object)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)

Issue #337, opened 2015-04-20
netmq high cpu utilization pub/sub socket


SriramSakthivel:
Hi,

I've lately migrated to netmq from clrzmq. I discovered that netmq consumes high cpu usage when using `SubscribeSocket`. 

I'm specifying the infinite receive timeout. It doesn't matter whether the data is received or not, netmq takes high cpu.

I've created a minimal reproduction code to demonstrate the problem.

Following is the code which I tested with clrzmq. Which works good; zero percentage cpu usage when waiting in `ReceiveFrame` method.

    namespace ClrzmqReceivePerformance
    {
        class Program
        {
            static void Main(string[] args)
            {
                Run();
            }

            private static void Run()
            {
                using (ZmqContext context = ZmqContext.Create())
                using (ZmqSocket publisher = context.CreateSocket(SocketType.PUB))
                {
                    publisher.Bind(string.Format("tcp://*:{0}", 8181));
                    //Publish nothing; Even if you publish, it makes no difference

                    using (ZmqSocket subscriber = context.CreateSocket(SocketType.SUB))
                    {
                        subscriber.Connect(string.Format("tcp://127.0.0.1:{0}", 8181));
                        subscriber.Subscribe(Encoding.Default.GetBytes(""));
                        subscriber.ReceiveTimeout = TimeSpan.MinValue;

                        while (true)
                        {
                            Frame received = subscriber.ReceiveFrame();
                            Console.WriteLine(Encoding.Default.GetString(received.Buffer, 0, received.BufferSize));
                        }
                    }
                }
            }
        }
    }

And the following code which uses netmq consumes high cpu time. In a single cpu machine this takes up all. 100 % utilization, 50 % in dual core machine and so on. 

    namespace NetmqReceivePerformance
    {
        class Program
        {
            static void Main(string[] args)
            {
                Run();
            }

            private static void Run()
            {
                using (NetMQContext context = NetMQContext.Create())
                using (NetMQSocket publisher = context.CreateSocket(ZmqSocketType.Pub))
                {
                    publisher.Bind(string.Format("tcp://*:{0}", 8181));
                    //Publish nothing; Even if you publish, it makes no difference

                    using (SubscriberSocket subscriber = (SubscriberSocket) context.CreateSocket(ZmqSocketType.Sub))
                    {
                        subscriber.Connect(string.Format("tcp://127.0.0.1:{0}", 8181));
                        subscriber.Subscribe(Encoding.Default.GetBytes(""));
[...]

PR #338, opened 2015-04-21
Fix for the issue #337

SriramSakthivel:
`Socket.Poll` should be called with the value of -1 timeout if you need a Infinite timeout. If you pass any other value the socket api may misbehave. When converting millisecond to microsecond we're not considering this. I've added a special case to address this issue.


drewnoakes:
Pushed as c0902f8e56a4f4c7c8e79d70daa7b5cc722e4f58.

Issue #344, opened 2015-05-15
Use of EventDelegatorHelper<T>


drewnoakes:
Just looking at this class, I can't see what it gives above just using a C# `event` (`MulticastDelegate`) directly. Am I missing something?

Also the code here looks suspicious:

* [NetMQActor constructor](https://github.com/zeromq/netmq/blob/master/src/NetMQ/NetMQActor.cs#L140)
* [Actor constructor](https://github.com/zeromq/netmq/blob/master/src/NetMQ/InProcActors/Actor.cs#L68) (deprecated)

Shouldn't the `unregisterFromEvent` actions be using `-=` instead of `+=`? Also, then 'send' variant uses different events in the register vs unregister actions.

I've looked at this for a while and can't work out if it's by design. If so, let's document it. If not, I'll get a PR together.

somdoron:
you are right about the bug in the code. We want to delegate an event from the NetMQSocket to the NetMQActor, however we only want to register once if anyone register to the NetMQActor and to unregister if not one register.

It is very small performance improvement (that might not needed, as usually only one register happen anyway), so if would not use it we would register to the origin event the number of times the event is register and we will create events args multiple times instead of once.

When coming to think of it, the improvement is very small and this only complicated the code. I'm ok with removing it.

PR #345, opened 2015-05-15
Bug fixes in use of EventDelegatorHelper<TArgs>

drewnoakes:
Following #344, here are some fixes to the use of internal `EventDelegatorHelper`.

1. Use correct events in all places
2. Fix resubscribe in place of unsubscribe
3. Fix bug in reference counting that prevents unsubscribe

In #344 we discussed removing `EventDelegatorHelper`. Looking at it a little further, it possibly makes a difference in [`NetMQSocket:275`](https://github.com/zeromq/netmq/blob/master/src/NetMQ/NetMQSocket.cs#L275) where the absence of a receive handler excludes the `PollEvents.PollIn` flag from being set.

@somdoron, if you still feel like removing this class, I will put together another PR. At least it seems that bugs are now fixed.

### DIFF
--- a/src/NetMQ.Tests/EventDelegatorTests.cs
+++ b/src/NetMQ.Tests/EventDelegatorTests.cs
@@ -1,0 +1,63 @@
+﻿using System;
+using NUnit.Framework;
+
+namespace NetMQ.Tests
+{
+    [TestFixture]
+    public class EventDelegatorTests
+    {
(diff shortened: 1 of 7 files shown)


drewnoakes:
Actually, hold off on merging this and I'll add some unit tests.

drewnoakes:
Ok, I think this is good to go if others agree.

PR #365, opened 2015-07-27
Convert Command from a class to a struct

drewnoakes:
Following some [discussion on the mailing list](https://groups.google.com/forum/#!topic/netmq-dev/sDuyucNPd2Y) in which brunobodin discovered allocations of `Command` objects when sending, I ran this experimental branch which converts `Command` to a value type, avoiding the heap allocation.

The change was fairly straightforward and completely within internal code, so no public API changes occurred.

I ran the `Performance\NetMQ.SimpleTests` project before and after this change, and the results look promising in terms of throughout and latency. See this screenshot of a diff.

image

Latencies and throughput are improved where message sizes are small. In some cases values are slightly worse, though I feel that's within the measurement error/uncertainty, so I'm not too concerned. Would be good to have someone else test this too.

### DIFF
--- a/src/NetMQ/Blob.cs
+++ b/src/NetMQ/Blob.cs
@@ -1,6 +1,6 @@
 /*
     Copyright (c) 2010 250bpm s.r.o.
-    Copyright (c) 2010-2011 Other contributors as noted in the AUTHORS file
+    Copyright (c) 2010-2015 Other contributors as noted in the AUTHORS file
 
     This file is part of 0MQ.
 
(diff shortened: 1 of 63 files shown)


somdoron:
I run the test, no big different before and after on my machine (actually is sometimes higher and some times lower).

Also the activate read is happen rearly (I think). Anyway the change is good

bbdb68:
Hi,

I am impressed, just two days away and the change is integrated... Thanks to Drew for coding the change and to Doron for integration. I will do my best to make use of it soon, but next days are rather busy for me.

Makes me feel that netmq/zeromq sounds as a good choice for my project !

Bruno

PR #366, opened 2015-07-27
C# 6

drewnoakes:
Now that C# 6 is out, is there any reason not to start taking advantage of its features? We can still target CLR versions 3.5 and 4.0.

* `nameof` operator (e0c5a4718cc2ea2404300f321fc276d4eb28f542)
* Null propagation operators (1903ff5223840a59d3e43c561ffddffad6e0f6a4)
* Immutable read only auto properties (f8afd9f0a2ba42c79e97a6ca9b9b654b948d301a)
* Expression bodies (aabf98da9000732f9d05422be04f73971c5a23c5)
* String interpolation (d5e382f83b78d4c6a9333b232fd9f850503d8400)

VS2015 Community Edition is available for open source work, so upgrading the project shouldn't exclude contributors.

I upgraded another of my projects recently to MSBuild 14.0 (VS2015/C#6) and had to [contact AppVeyor](http://help.appveyor.com/discussions/problems/2534-building-vs2015-rtm-solution) to get the project migrated to a build server that worked. I'm happy to sort that out.

Please share any thoughts here. Each of the C# 6 features has been introduced in a separate commit, so we can pick the PR apart if we want to take only some of the features and not others.

Do people like these new language features for NetMQ?

### DIFF
--- a/src/NetMQ.Tests/EventDelegatorTests.cs
+++ b/src/NetMQ.Tests/EventDelegatorTests.cs
@@ -13,7 +13,7 @@ namespace NetMQ.Tests
                 Value = value;
             }
 
-            public T Value { get; private set; }
+            public T Value { get; }
         }
 
         private event EventHandler<Args<int>> Source;
--- a/src/NetMQ.Tests/InProcActors/AccountJSON/AccountShimHandler.cs
+++ b/src/NetMQ.Tests/InProcActors/AccountJSON/AccountShimHandler.cs
@@ -65,7 +65,7 @@ namespace NetMQ.Tests.InProcActors.AccountJSON
                 // demonstration purposes only, any unhandled fault will bubble up to caller's code.
                 catch (Exception e)
                 {
-                    shim.SendFrame(string.Format("Error: Exception occurred {0}", e.Message));
+                    shim.SendFrame($"Error: Exception occurred {e.Message}");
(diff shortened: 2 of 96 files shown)


somdoron:
I really like it all of it... mostly the string interpolation.

Anyway, the think I wonder if people will not contribute if they will have to download a new version of visual studio. two month from now it probably won't be a problem. What do you think?

drewnoakes:
I don't know how fast adoption rates will be. It's a valid concern.

These changes were pretty easy to apply using ReSharper. We can hold off
and reapply later if this PR doesn't merge at that time.

I'll split the non-C# 6 changes out to a new PR.

JamesWHurst:
I honestly don't know - but my guess would be that most project-contributors are pretty energetic about jumping on to the latest tool version. I like the C# enhancements, and note that you can still target .NET 4 with them. I'd say go for it. Just be cognizant that we have a .NET 3.5 version that needs to be maintained (it's still 3.5 compatible, right?).

drewnoakes:
I believe the .NET 3.5 project still targets correctly, yes.

drewnoakes:
Just realised our build is with TeamCity, not AppVeyor.

drewnoakes:
Too hard to rebase this for now. Closing.

PR #377, opened 2015-09-02
Prepare version 3.3.2.0

somdoron:
I changed the assembly info version and release notes.

Once merged I will release it to nuget.

@drewnoakes what do you think?


drewnoakes:
Looks great!

PR #395, opened 2015-10-26
problem: no easy way to reset timer

somdoron:
The only the reset the timer interval is to disable and enable the timer.

Add a method called EnableAndReset that enable the timer if disabled and reset the interval.

Useful Example: when you use the timer to send heartbeat after you didn't get a message for X seconds. Each time you get a message you will just call EnableAndReset.

### DIFF
--- a/src/NetMQ.Tests/PollerTests.cs
+++ b/src/NetMQ.Tests/PollerTests.cs
@@ -655,6 +655,57 @@ namespace NetMQ.Tests
         }
 
         [Test]
+        public void ResetTimer()
+        {
+            using (NetMQContext context = NetMQContext.Create())
+            {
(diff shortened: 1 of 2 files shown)


drewnoakes:
Looks like there's a compilation error in the .NET 3.5 project:

>  src\NetMQ.Tests\NetMQ3.5.Tests.csproj
>
> PollerTests.cs(674, 44): error CS0246: The type or namespace name 'NetMQScheduler' could not be found (are you missing a using directive or an assembly reference?)

somdoron:
I'm using the scheduler for test... I will try and fix that.

somdoron:
@drewnoakes it is working now...

PR #397, opened 2015-10-29
Allow reconnect on access denied socket error

ondrejtomcik:
This change should handle a case, when company has a different policy on wifi and ethernet network. If specific connection is allowed on ethernet, but on wifi it's blocked by firewall outgoing rule policy, it generates access denied socket error. In this case reconnection should be allowed to establish connection again when user is back on ethernet - has set policy which allows it.


ondrejtomcik:
@somdoron when I can expect updated nuget package please ? Thanks

somdoron:
@ondrejtomcik I will try to release today, but it will be under beta for a while

ondrejtomcik:
Sure, thank you.

somdoron:
@ondrejtomcik you can find a new beta version on nuget:
https://www.nuget.org/packages/NetMQ/3.3.2.3-beta

PR #400, opened 2015-11-10
Add offset to Msg to support ArraySegment-like API with `byte[], offset, length`

buybackoff:
This is for discussion at this stage - to ask what do you think about such a change?

I have noticed that most extensions for sending data make additional copy of data even if it is supplied as `byte[]`. But even in the lowest level `Send(ref msg)` and in Msg implementation, currently there is no support for `offset` of a `byte[]`, only its `length`.  However, `byte[] data, int offset, int length` is the standard pattern used almost everywhere in .NET. `ArraySegment<byte>` is an equivalent to the trio and is just a struct around `byte[]` with two fields, it is used in MSFT API with WebSockets. Without offsets, we are forced to copy arrays one more time if payload starts at not zero (which is not a rare case when the whole system is designed for zero-copy when possible).

It looks like initially NetMQ was designed after ZeroMQ, but in C/C++ they work with pointers + length, any offset could be applied by pointer arithmetic. In .NET world, if a payload is placed within a buffer at some offset, we have to copy it to use the initial NetMQ APIs that had ignored offsets. With this commit, Msg supports an offset property and behaves mostly like ArraySegment. 

All tests pass. Performance impact is non-conclusive based on SimpleTests project. Need more evaluation, mostly depends on the use case - if input data is mostly like ArraySegment<byte>/buffers with non-zero offsets then zero-copy could be beneficial, especially for IPC.

Next step could be to add implicit conversion to/from `ArraySegment<byte>/Msg` and add convenient extensions for `ArraySegment<byte>`.

### DIFF
--- a/src/NetMQ/Msg.cs
+++ b/src/NetMQ/Msg.cs
@@ -96,7 +96,13 @@ namespace NetMQ
         /// <summary>
         /// Get the number of bytes within the Data property.
         /// </summary>
-        public int Size { get; private set; }
+        public int Count { get; private set; }
+
+        /// <summary>
+        /// Gets the position of the first element in the Data property delimited by the message,
+        //     relative to the start of the original array.
+        /// </summary>
+        public int Offset { get; private set; }
 
         #region MsgType
 
@@ -184,7 +190,7 @@ namespace NetMQ
         /// This value will be <c>null</c> if <see cref="MsgType"/> is <see cref="NetMQ.MsgType.Uninitialised"/>,
         /// <see cref="NetMQ.MsgType.Empty"/> or <see cref="NetMQ.MsgType.Delimiter"/>.
         /// </remarks>
-        public byte[] Data { get; private set; }
+        public byte[] Array { get; private set; }
 
         /// <summary>Get whether this <see cref="Msg"/> is initialised and ready for use.</summary>
         [Obsolete("Use the IsInitialised property instead")]
@@ -202,8 +208,9 @@ namespace NetMQ
         {
             MsgType = MsgType.Empty;
             Flags = MsgFlags.None;
-            Size = 0;
-            Data = null;
+            Count = 0;
+            Offset = 0;
+            Array = null;
             m_refCount = null;
         }
 
@@ -215,8 +222,9 @@ namespace NetMQ
         {
             MsgType = MsgType.Pool;
             Flags = MsgFlags.None;
-            Data = BufferPool.Take(size);
-            Size = size;
+            Array = BufferPool.Take(size);
+            Count = size;
+            Offset = 0;
             m_refCount = new AtomicCounter();
         }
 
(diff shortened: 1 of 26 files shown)


somdoron:
In general I think it is a good idea, I want to make sure the PR is not breaking anything and can merge it.

src/NetMQ/Msg.cs
```diff
@@ -225,15 +233,17 @@ public void InitPool(int size)
         /// </summary>
         /// <param name="data">the byte-array of data to assign to the Msg's Data property</param>
         /// <param name="size">the number of bytes that are in the data byte-array</param>
-        public void InitGC([NotNull] byte[] data, int size)
+        public void InitGC([NotNull] byte[] data, int size, int offset = 0)
```
somdoron: I think it is better to create a new constructor here, first this can break if you are not recompiling.
Also it would be better to keep .net structure of array,offset,length.

somdoron:
Looks good, I left you one comment in the code.

Also can you make a test that use the Msg with the new offset field?

buybackoff:
Will do. Also what do you recommend for the case when input byte array is from a pool shared with NetMQ pool (e.g. an application buffer pool implementing NetMQ pool interface)? I have though about a method like `InitWrap` that takes an existing array an treats it as pooled. But there are two situations: the first is when NetMQ takes ownership of a buffer and must then put it back to the pool, and another is when NetMQ borrows a buffer and should not return it to the pool. Haven't figured out the correct API yet.

somdoron:
I think the only way to create a Msg from buffer pool is from InitPool. So if you take the buffer directly from the pool (Which I don't recommend) and use it with Msg it won't be returned.

buybackoff:
I mean the case when the buffer pool implementation is shared between my app and NetMQ, my app takes a buffer before touching NetMQ, writes data to it and then creates a message of it. In Rust-like terminology I think methods like `InitOwn` and `InitBorrow` could be used. However, `InitGC` behaves exactly like `InitBorrow`, and since all `Send` operations are synchronous and blocking, we could always return a buffer explicitly after sending. In NetMQ's terminology we could add a method `InitMove` that behaves like `InitOwn` and `Msg.Move`. However, it could only be used with a buffer taken from a pool before, but in other cases we will be returning buffer that was not taken from a pool. It could be a convenient shortcut, but anyone who would know why to use it could just as easily return a buffer manually after sending.

The bottom line is that `InitGC` + buffer management outside NetMQ is probably enough. It would be nice to have a short documentation about zero-copy and integration with external buffer pools with issues like above.

somdoron:
Send is not synchronous. After calling Send you are not allow to change the buffer. Also the Msg object return from Send operation is empty (not attached to any buffer).

Regarding you usage, Instead of taking a buffer from the Pool create Msg object, init it from Pool with the correct size and use that instead.

So the Send operations in NetMQ is tuned for zero-allocation but not zero-copy, the problem is with the Receive operation which both allocate and copy data. (at least with the byte array overloads).

buybackoff:
What kind of test should I add? I have run a perf test with offsets [here](https://github.com/buybackoff/netmq/blob/839902ef3e2cdf26ce1f717f870268330fc3f7cb/src/RouterPerformance/Program.cs#L45). Should I add a snippet from there to tests?

somdoron:
I merging this. Anyway none of the tests test the Msg with offset over the wire, to make sure the encoders actually encode and decode it correctly. So if you can add a test with TCP sockets that send a Msg with offset that will be great. If not I will add it later.

buybackoff:
Great, thanks! All existing users have offset 0 everywhere, so it kind of safe for now. Will add a test that ensures working offset over TCP later.

PR #401, opened 2015-11-15
Broadcast feature for XPub socket

buybackoff:
These changes add a broadcast feature that is similar to SignalR's `Clients.Others`. If the broadcast option is enabled and an XSub socket sends a topic prefixed with `2`, then that socket will never be matched on a subsequent `Send` by XPub. With this feature it is easy to turn a XPub into a message bus, when N XSub clients send a message to every other one - without additional Dealer/Router socket and without complicated check if a sender receives its own message back.

Without `XPubBroadcast` set to true, there is no change in behavior, so this change is backward compatible.

This PR depends on #400. It also removes unneeded copying of data in XSub by using the offset feature from #400.

### DIFF
--- a/src/NetMQ.Tests/RouterTests.cs
+++ b/src/NetMQ.Tests/RouterTests.cs
@@ -2,11 +2,14 @@
 using NUnit.Framework;
 using System.Text;
 
-namespace NetMQ.Tests {
+namespace NetMQ.Tests 
+{
     [TestFixture]
-    public class RouterTests {
+    public class RouterTests 
+    {
         [Test]
-        public void Mandatory() {
+        public void Mandatory() 
+        {
             using (var context = NetMQContext.Create())
             using (var router = context.CreateRouterSocket()) {
                 router.Options.RouterMandatory = true;
@@ -17,7 +20,8 @@ namespace NetMQ.Tests {
         }
 
         [Test]
-        public void ReceiveReadyDot35Bug() {
+        public void ReceiveReadyDot35Bug() 
+        {
             // In .NET 3.5, we saw an issue where ReceiveReady would be raised every second despite nothing being received
 
             using (var context = NetMQContext.Create())
@@ -33,7 +37,8 @@ namespace NetMQ.Tests {
 
 
         [Test]
-        public void TwoMessagesFromRouterToDealer() {
+        public void TwoMessagesFromRouterToDealer() 
+        {
             using (var context = NetMQContext.Create())
             using (var poller = new Poller())
             using (var server = context.CreateRouterSocket())
--- a/src/NetMQ/Core/Patterns/XPub.cs
+++ b/src/NetMQ/Core/Patterns/XPub.cs
@@ -59,7 +59,10 @@ namespace NetMQ.Core.Patterns
         /// </summary>
         private bool m_manual;
 
+        private bool m_broadcastEnabled;
+
         private Pipe m_lastPipe;
+        private bool m_lastPipeIsBroadcast;
 
         private Msg m_welcomeMessage;
 
@@ -82,7 +85,11 @@ namespace NetMQ.Core.Patterns
             s_markAsMatching = (pipe, data, size, arg) =>
             {
                 var self = (XPub)arg;
-                self.m_distribution.Match(pipe);
+                // skip the sender of a broadcast message
+                if (!(self.m_broadcastEnabled && self.m_lastPipeIsBroadcast && self.m_lastPipe == pipe))
+                {
+                    self.m_distribution.Match(pipe);
+                }
             };
 
             s_sendUnsubscription = (pipe, data, size, arg) =>
@@ -157,6 +164,7 @@ namespace NetMQ.Core.Patterns
             var sub = new Msg();
             while (pipe.Read(ref sub))
             {
+                m_lastPipeIsBroadcast = false;
                 // Apply the subscription to the trie.
                 int size = sub.Size;
                 if (size > 0 && (sub[0] == 0 || sub[0] == 1))
@@ -179,12 +187,20 @@ namespace NetMQ.Core.Patterns
                             m_pending.Enqueue(sub.CloneData());
                     }
                 }
+                else if (m_broadcastEnabled && size > 0 && (sub[0] == 2))
+                {
+                    m_lastPipe = pipe;
+                    m_lastPipeIsBroadcast = true;
+                    sub.Offset = sub.Offset + 1;
+                    sub.Size = sub.Size - 1;
+                    XSend(ref sub);
+                }
                 else // process message unrelated to sub/unsub
                 {
                     m_pending.Enqueue(sub.CloneData());
                 }
 
-                sub.Close();
+                if(!m_lastPipeIsBroadcast) sub.Close();
             }
         }
 
@@ -219,6 +235,11 @@ namespace NetMQ.Core.Patterns
                     m_manual = true;
                     return true;
                 }
+                case ZmqSocketOption.XPublisherBroadcast:
+                    {
+                        m_broadcastEnabled = true;
+                        return true;
+                    }
                 case ZmqSocketOption.Subscribe:
                 {
                     if (m_manual && m_lastPipe != null)
--- a/src/NetMQ/Core/Patterns/XPub.cs
+++ b/src/NetMQ/Core/Patterns/XPub.cs
@@ -164,7 +164,6 @@ namespace NetMQ.Core.Patterns
             var sub = new Msg();
             while (pipe.Read(ref sub))
             {
-                m_lastPipeIsBroadcast = false;
                 // Apply the subscription to the trie.
                 int size = sub.Size;
                 if (size > 0 && (sub[0] == 0 || sub[0] == 1))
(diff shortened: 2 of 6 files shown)


src/NetMQ.Tests/RouterTests.cs
```diff
@@ -1,17 +1,14 @@
 ﻿using System;
 using NUnit.Framework;
+using System.Text;
 
-namespace NetMQ.Tests
-{
+namespace NetMQ.Tests {
```
somdoron: This change the style of the project. Can you fix this?

buybackoff: Sure! Actually that test was not related to the PR, but to another issue I asked about. I accidentally added it.

somdoron:
Very nice. I think you have a bug, after a subscriber send a broadcast message this subscriber won't receive regular publisher messages. You should set the m_lastPipeIsBroadcast to false after calling XSend (it also good idea to nullify the m_lastPipe to avoid keeping a reference to terminated pipe.

buybackoff:
It is set to false when there is no more message parts here: https://github.com/buybackoff/netmq/blob/feature_broadcast/src/NetMQ/Core/Patterns/XPub.cs#L325

src/NetMQ/Core/Patterns/XPub.cs
```diff
@@ -158,34 +166,41 @@ protected override void XReadActivated(Pipe pipe)
...
+                    m_pending.Enqueue(sub);
+                    //XSend(ref sub);
+                }
                 else // process message unrelated to sub/unsub
                 {
-                    m_pending.Enqueue(sub.CloneData());
+                    m_pending.Enqueue(sub);
                 }
 
-                sub.Close();
```
somdoron: why you are not closing the Msg? there is cases the messages should be closed (it is not always enqueued)

src/NetMQ/Core/Patterns/XPub.cs
```diff
@@ -323,13 +346,8 @@ protected override bool XRecv(ref Msg msg)
             if (m_pending.Count == 0)
                 return false;
 
-            msg.Close();
```
somdoron: This should be called.

somdoron:
Also can you rebased it on master so I will only display relevant commits? As this is a big change I want to make sure it doesn't break anything

buybackoff:
An implicit requirement is that a broadcast message is sent from withing publisher's ReceiveReady event. Given that poller is single-threaded, is it possible that a regular send could call XSend after a publisher receives a message but before the publisher sends it inside the event?

```
private void _publisher_ReceiveReady(object sender, NetMQSocketEventArgs e)
        {
            var socket = e.Socket;
            var id = new Msg();
            id.InitEmpty();
            socket.Receive(ref id);
            Debug.Assert(id.HasMore);
            var request = new Msg();
            request.InitEmpty();
            socket.Receive(ref request);
             
             // could regular pub call XSend here? They are inside a single thread, it should not be possible

            // resend a message to all clients that are subscribed to the id, except the sender
            _publisher.Send(ref id, true);
            _publisher.Send(ref request, false);
        }
```

src/NetMQ/Core/Patterns/XPub.cs
buybackoff: Only if this if condition is false: https://github.com/buybackoff/netmq/blob/feature_broadcast/src/NetMQ/Core/Patterns/XPub.cs#L186

will fix this

src/NetMQ/Core/Patterns/XPub.cs
buybackoff: Agree. But shouldn't we check if it shared? Ot it is never shared at this point?

src/NetMQ/Core/Patterns/XPub.cs
```diff
@@ -176,15 +184,29 @@ protected override void XReadActivated(Pipe pipe)
...
+                            m_pending.Enqueue(sub);
+                        }
+                        else
+                        {
+                            sub.Close();
+                        }
+
                     }
                 }
+                else if (m_broadcastEnabled && size > 0 && sub.HasMore && sub[0] == 2)
```
somdoron: why do you force the HasMore?

somdoron:
It is not possible...

src/NetMQ/Core/Patterns/XPub.cs
buybackoff: because only the first part with the topic is prefixed. The second part, in general, could have 2 as the first byte.

buybackoff: I think that subscribe/unsubscribe should also work only with the first message. Now if the second message with payload starts with 0/1, a weird thing will happen.

src/NetMQ/Core/Patterns/XPub.cs
```diff
@@ -176,15 +184,29 @@ protected override void XReadActivated(Pipe pipe)
...
+                {
+                    m_lastPipe = pipe;
+                    m_lastPipeIsBroadcast = true;
+                    sub.Offset = sub.Offset + 1;
+                    sub.Size = sub.Size - 1;
+                    m_pending.Enqueue(sub);
```
somdoron: Should we keep the prefix so it behave like other messages? like sub and unsub and the rest.

That way how can we tell if it a broadcast message, subscribe, unsubscribe or something else?

src/NetMQ/Core/Patterns/XPub.cs
somdoron: So that is not the correct way to check if this is the first frame. What if you have 3 frames? what if it is one frame?

So current NetMQ implementation doesn't support multi-messages from xsub to xpub (ZeroMQ is actually do supporting). Anyway I suggest we keep at the beginning for 1 frame and later add support to more frames.

src/NetMQ/Core/Patterns/XPub.cs
buybackoff: Regular subscribe/unsubscribe do not have a second frame. Regular send does not add a prefix to a topic. So in regular send, the first frame is a topic, and the second frame is a payload. In the case of the broadcast, topic is always prefixed with `2`, but subscribers should not care if a message is a regular one or a broadcasted one. We only exclude the sender in case of a broadcast message.

buybackoff: If we keep the `2` prefix here, Match won't work and the subsequent logic will be very complicated.

somdoron: PubSub can also work with one frame only... the first part being a topic is not in the code or RFC. just convenient.

Anyway I'm talking about the XPub receive...

So before this change, if you call XPub receive you can get:
If it a subscribe it will be prefix with 0, ubsubscribe with 1 and the rest with something else

Now you actually get a message that you cannot tell what type is it, because you removed the prefix.

The user code that call for Send should remove the prefix.

buybackoff: But I remove prefix only when `sub[0] == 2`, so it doesn't affect sub/unsub messages. The only thing it does is marking the sender (last pipe) for exclusion, but the prefix is not needed for anything else. After these lines, the message should look like a regular one to everyone. The sender just won't receive it back.

somdoron: Still, what if the next byte is 1 and the user code will think it's a subscription? Do you want to skype over this?

buybackoff: This is a good point! But the we need to change the match logic in XSend so that it ignores the first byte in the first frame. I will remove the check for .HasMore to support a single frame case. And will change the match logic.

somdoron: Wait, don't change the match logic. It will affect too many areas. The should just remove the prefix before calling the send.

So the user check the prefix byte, if it is 2, remove the prefix and send the message.

buybackoff: Do you mean that we should change topic here:
```
private void _publisher_ReceiveReady(object sender, NetMQSocketEventArgs e)
        {
            var socket = e.Socket;
            var id = new Msg();
            id.InitEmpty();
            socket.Receive(ref id);
            Debug.Assert(id.HasMore);
            var request = new Msg();
            request.InitEmpty();
            socket.Receive(ref request);

             // Remove the first byte if it is 2

            // resend a message to all clients that are subscribed to the id, except the sender
            _publisher.Send(ref id, true);
            _publisher.Send(ref request, false);
        }
```

somdoron: yes. We can also add a method to Msg to do this quickly. Like TrimPrefix.

buybackoff: It is trivial with these lines
```
sub.Offset = sub.Offset + 1;
sub.Size = sub.Size - 1;
```
I have committed the change and updated the test.

somdoron: It is internal, but we can fix this later if needed.

buybackoff:
Yay! :smile:  Now the clone pattern in NetMQ could be achieved much simpler than with several sockets like shown in the Guide.

somdoron:
Great contribution, thanks.
So go ahead and write clone example with this new feature and send a PR :-)

Also consider writing a documentation for this on our documentation page.

PR #413, opened 2015-12-23
Added queue IENumerator and IENumerable methods

gitn00b1337:


somdoron:
Your implementation is not thread safe. Implement only IEnumerable and use yield return

gitn00b1337:
Ok I've implemented a commit again.. do i have to make a new pull request or can i update this somehow

gitn00b1337:
in fact i noticed you commented on that one. At this point if I'm still wrong just cyber slap me and do the change yourself... i'll be here all night trying haha

PR #421, opened 2016-01-08
Add NetMQContextExtensions for encapsulation of Retry

dalebrubaker:
I'm rather new to both NetMQ (and GitHub -- my first pull request). So I really have no idea what to expect. :)

My motivation was to simplify the use of Lazy Pirate. It seems rather common to want to make a request with retry. This is my attempt at encapsulation.

I've added a couple of sample extensions to a class NetMQContextExtensions
    RequestResponseMultipartMessageWithRetry()
    RequestResponseStringWithRetry()

Then I've made a Samples.PiratePattern.LazyPirate.Client2 program that shows the dramatic reduction in code whenever request-with-retry is desired.

Unit tests are included.

I haven't added this to NetMQ3.5. I thought I'd wait to see what happens here.

Best regards,
Dale


somdoron:
Nice. Two comments:

* I'm not sure it should be an extension to the context (it somehow not very related), can it be a RequestUtility class? or static method on RequestSocket class?

* I'm not a big fun of the progress interface and class, you can just provide a push/pub socket instead and the progress will be send to the socket (if not null), more elegant.

Let me know what you think, I will merge it anyway.

dalebrubaker:
No rush. I share both your concerns.
 1. I'll start a RequestUtility class. Static?
 2. I love the push/pub socket idea, but I'm lost (too new to NetMQ). Can you point me to an example or give me some big hints?

somdoron:
1. I think static is fine for utility class, anyway what ever you prefer.

2. Just accept PublisherSocket instead of the Progress and call send. Take  a look at the pubsub example in the docs.

dalebrubaker:
I went with your idea about a static method on RequestSocket. And I changed away from IProgress<T> to pub/sub, which is really cool.
The files are changed, but I'm happy to remove the context arguments from the method signatures if you get your changes merged before merging mine.
Thanks.

somdoron:
We can change that later. 

There is a compilation error:

```
Compilation error: src\Samples\Pirate Pattern\Lazy Pirate\LazyPirate.Server2\LazyPirate.Server2.csproj 
Open in IDE
CSC error CS2001: Source file 'Program.cs' could not be found
```

dalebrubaker:
Sorry. Fixed.

somdoron:
merged. You can also send another PR without the context...

PR #423, opened 2016-01-09
Problem: context must be used to create sockets

somdoron:
Suggested solution: global context, now sockets can be created without a context.

### DIFF
--- a/src/NetMQ/Global.cs
+++ b/src/NetMQ/Global.cs
@@ -1,0 +1,114 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using NetMQ.Core;
+
+namespace NetMQ
+{
+    public static class Global
+    {
+        private static int s_ioThreadCount = 1;
+        private static int s_maxSockets = 1024;
+
+        private static Ctx s_ctx;
+        private static object s_sync = new object();
+        private static int s_socketCount = 0;
+
+        /// <summary>
+        /// Get or set the number of IO Threads in the context, default is 1.
+        /// 1 is good for most cases.
+        /// </summary>
+        public static int ThreadPoolSize
+        {
+            get
+            {
+                lock (s_ctx)
+                {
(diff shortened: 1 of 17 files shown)


dalebrubaker:
Seems too good to be true. :) Is there still a purpose for context? (In other words, what does this really mean?) Thanks.

src/NetMQ/Global.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+
+        internal static SocketBase CreateSocket(ZmqSocketType socketType)
+        {
+            lock (s_sync)
+            {
+                if (s_ctx == null)
+                {
+                    s_ctx = new Ctx();
```
dalebrubaker: So essentially this is a singleton, and this "guarantees" there is only one context per process, but it's handled behind the scene. No need for explicit NetMQContext unless more than one per process was needed for some reason. Do I understand correctly? Thanks.

somdoron:
Context are old concepts from zeromq, actually there is almost no use for them in NetMQ.

One problem is that now when you dispose a socket it doesn't really get disposed immediately, it queue for termination on a dedicate thread. So the only way to make sure all sockets are disposed is to call Context.Dispose. I'm going to change it so when you call Socket.Dispose it will block until the socket is disposed. However this can break some application so I need to find a way to do it without breaking anything.

src/NetMQ/Global.cs
somdoron: Yes.

drewnoakes:
Looks good Doron. Is this ready for merging? Would it be worth marking `NetMQContext` as obsolete too?

somdoron:
Yes it is ready. Tomorrow I will work on make the socket dispose itself completely, currently the last disposed socket is terminate the context, however this can take time, I prefer that each socket will get disposed completely when dispose is called . Regarding making the NetMQContext obsolete, let's make this stable first and then make it obsolete.

PR #426, opened 2016-01-10
Removed NetMqContext references from new RequestSocket.RequestWithRet…

dalebrubaker:
...ry Methods

These changes have been made, but Unit Tests fail when run together, some of the time.
My unit tests run individually, but sometimes one of them will fail when they are run as a batch. And sometimes the NUnit test runner is crashing (always crashes when I try to Run All unit tests in the solution).
And I sometimes get a Debug.Assert error at line 59 of NetMQ.Core.ErrorHelper.
   the asserted argument error is "SocketError"
stack trace:
  at AsyncIO.Windows.Socket.Bind(IPEndPoint localEndPoint)
   at NetMQ.Core.Transports.Tcp.TcpListener.SetAddress(String addr) in C:\GitDev\NetMQ\src\NetMQ\Core\Transports\Tcp\TcpListener.cs:line 140
   
Lazy Pirate samples run, but they seem "jerky". Not a technical term, and maybe my imagination.
I don't know if this is related to the conversation you are having about socket disposal.
So I haven't dug deeply into this. Let me know if you think I should do something...


dalebrubaker:
The Team City log gave me some more info. It appears that I have written the unit tests so that they can interfere with each other -- multiple tests trying to reconnect simultaneously to the same address. Hmm. I'll change that.

somdoron:
So two options

* are you sure you disposed the router socket?
* because context is not disposed between tests the socket might not get disposed completely before next test. You can use socket.BindRandomPort  which use free random port and return the used port.

src/NetMQ.Tests/RequestWithRetryTests.cs
```diff
@@ -20,13 +20,12 @@ public void RequestResponseMultipartMessageWithRetrySucceedsFirstTry()
             var requestTimeout = TimeSpan.FromMilliseconds(100);
             var requestMessage = new NetMQMessage(1);
             requestMessage.Append("Hi");
-            using (var context = NetMQContext.Create())
-            using (var progressPublisher = context.CreatePublisherSocket())
+            using (var progressPublisher = new PublisherSocket())
             {
                 progressPublisher.Bind("tcp://127.0.0.1:5556");
                 Task.Factory.StartNew(() =>
```
somdoron: this loops never ends...

src/NetMQ.Tests/RequestWithRetryTests.cs
```diff
@@ -39,7 +38,7 @@ public void RequestResponseMultipartMessageWithRetrySucceedsFirstTry()
                 });
                 Task.Factory.StartNew(() =>
                 {
-                    using (var rep = context.CreateResponseSocket())
+                    using (var rep = new ResponseSocket())
                     {
                         rep.Bind(address);
                         while (true)
```
somdoron: this thread never ends

somdoron:
I think you have some endless threads in the tests

dalebrubaker:
Yes, thanks. Working on it now.

dalebrubaker:
Here is the assertion I get sometimes when I run all of the RequestWithRetryTests at the same time:
image

somdoron:
In which test?

dalebrubaker:
Run all 9 at a time: (RequestWithRetryTests)
1. All succeed.
2. All succeed, get an assertion that goes away before I could read it.
3. 5 succeed, 4 fail
image
4. 7 succeed, 2 fail:
image
5. Same exactly as the last run (7 succeed, 2 fail)
6. 8 succeed, 1 fails:
image

If there is a common them, it seems that it only occurs when I retry (TerminateSocket() then GetNewRequestSocket() in the RequestSockets class).

dalebrubaker:
But the Team City test runner found 4 failures and two of them were FirstTry.

somdoron:
This usually happens when two different threads accessing same socket. I
will take a look tomorrow

somdoron:
@dalebrubaker your threads still never exist. Anyway I'm merging this and will fix the tests.

PR #427, opened 2016-01-11
problem: creating socket in NetMQ is verbose and require multiple of lines

somdoron:
like bind and connect methods, which can be multiple lines of code.

Solution: accept connection string parameter in the ctor. For example:

```csharp
/// <summary>
/// Create a new DealerSocket and attach socket to zero or more endpoints.               
/// </summary>                
/// <param name="connectionString">List of NetMQ endpoints, seperated by commas and prefixed by '@' (to bind the socket) or '>' (to connect the socket).
/// Default action is connect (if endpoint doesn't start with '@' or '>')</param>
/// <example><code>var socket = new DealerSocket(">tcp://127.0.0.1:5555,@127.0.0.1:55556");</code></example>                 
public DealerSocket(string connectionString = null)
```

### DIFF
--- a/src/NetMQ.Tests/SocketTests.cs
+++ b/src/NetMQ.Tests/SocketTests.cs
@@ -787,18 +787,90 @@ namespace NetMQ.Tests
         [Test]
         public void ContextFree()
         {
-            using (DealerSocket server = new DealerSocket())
+            using (DealerSocket server = new DealerSocket("@tcp://127.0.0.1:5555"))
             {
-                using (DealerSocket client = new DealerSocket())
-                {
-                    server.Bind("tcp://127.0.0.1:5555");
-                    client.Connect("tcp://127.0.0.1:5555");
-
+                using (DealerSocket client = new DealerSocket(">tcp://127.0.0.1:5555"))
+                {                    
                     client.SendFrame("Hello");
 
                     Assert.AreEqual("Hello", server.ReceiveFrameString());
                 }
             }            
         }
+
+        [Test]
+        public void ConnectionStringDefault()
+        {
+            using (ResponseSocket response = new ResponseSocket("tcp://127.0.0.1:5555"))
+            {
+                using (RequestSocket request = new RequestSocket("tcp://127.0.0.1:5555"))
+                {
+                    request.SendFrame("Hello");
+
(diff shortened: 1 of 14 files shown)


drewnoakes:
Very cool to support this via the ctor.

Would it be clearer to have an API like:

```csharp
public DealerSocket(string connect = null, string bind = null)
```

Then callers:

```csharp
new DealerSocket("tcp://asdf:1234") // connect (implicit)
new DealerSocket(connect: "tcp://asdf:1234") // connect (explicit)
new DealerSocket(bind: "tcp://asdf:1234") // bind (explicit)
```

You could still comma separate lists of addresses for connect.

somdoron:
So two reason I did it like this:
* Same as CZMQ
* We can develop connection string DSL, so you can store connection string in config file. Maybe even add a way to configure some of the options like "@tcp://localhost:1234,#linger=0"

We can make another overload that accept your suggestion.

Anyway NetMQ become less verbose:

```csharp
  using (RequestSocket client = new RequestSocket("@tcp://127.0.0.1:5555"))
            using (ResponseSocket server = new ResponseSocket(">tcp://127.0.0.1:5555, @tcp://127.0.0.1:5556"))
            using (RequestSocket client2 = new RequestSocket("tcp://127.0.0.1:5556"))
            {
                client.SendFrame("Hello");
                Console.WriteLine(server.ReceiveFrameString());

                server.SendFrame("Hello back");
                Console.WriteLine(client.ReceiveFrameString());

                client2.SendFrame("Hello");
                Console.WriteLine(server.ReceiveFrameString());

                server.SendFrame("Hello back");
                Console.WriteLine(client2.ReceiveFrameString());
            }
```

drewnoakes:
If there's a precedent with CZMQ, then let's leave it. I can't see how an
overload would work. I'll get used to the two characters -- it's not too
magical.

Great seeing all this activity on the project over the last few days. I was
offline for a while and actually did some work on the merging of Poller and
NetMQScheduler. I hope it rebases properly now :) But being able to skip
using the context will actually make the patch much nicer.

somdoron:
great news. So after NetMQPoller we can make a clone repository of NetMQ called NetMQ3, make this one NetMQ 4 and remove all obsolete code :-)

PR #428, opened 2016-01-11
Problem: when wanting to quickly handle messages on new thread a lot of code have to be written

somdoron:
solution: NetMQProactor, small class to handle messages of one socket on different thread.

Also 	fixes RequestWithRetryTests and add context less method to NetMQActor.


somdoron:
@dalebrubaker this fix the problem with the tests

dalebrubaker:
@somdoron That is beautiful code! Thanks.

PR #436, opened 2016-01-14
Fix test and add some more docs

drewnoakes:
This should fix the build. I've also added some stub docs for `NetMQTimer`, `NetMQQueue` and `NetMQProactor`.

### DIFF
--- a/src/NetMQ/EmptyArray.cs
+++ b/src/NetMQ/EmptyArray.cs
@@ -1,0 +1,7 @@
+namespace NetMQ
+{
+    internal static class EmptyArray<T>
+    {
+        public static readonly T[] Instance = new T[0];
+    }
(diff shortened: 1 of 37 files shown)


src/NetMQ/EmptyArray.cs
```diff
@@ -0,0 +1,7 @@
+namespace NetMQ
+{
+    internal static class EmptyArray<T>
+    {
+        public static readonly T[] Instance = new T[0];
```
somdoron: nice

somdoron: should this be under core\utils?

drewnoakes: Probably belongs there, yes.

PR #442, opened 2016-01-20
Beacon sample

dalebrubaker:
Again, I don't know if this is useful or helpful. I did it for my own education.

I'm very tempted to work on a NetMQ version of zyre/jyre, for a project I'm working on.
Would you encourage that, or discourage it?
And is it necessarily separate from NetMQ, or could it be part of it? (It seems tiny.)

Best regards,
Dale


somdoron:
Great. Always welcome.

Also if you want to implement zyre check out  zproto for NetMQ, I making a video on how to use which will be available soon.

https://www.nuget.org/packages/NetMQ.zproto/

Regarding the correct location for zyre for NetMQ, I actually not sure, first I think every project should have repository, however it is hard to build a community and we actually have very good one around NetMQ which can make a good use of zyre for .Net.

Anyway it should be separate project, it may be under the NetMQ repository or different one. 

Bottom line, this is up to you.

PR #447, opened 2016-01-24
removed JetBrains.Anotations

awb99:
and added Compiled AsyncIO into lib folder


awb99:
As to AsyncIO project: One can tell github to include the full sourcecode of another project when the pull is done. I have seen this onother projects in the past. Unfortunatley Idont know how to do it myself.

drewnoakes:
Hi. Could you explain the rationale for removing `JetBrains.Annotations`, and also for not using NuGet for the AsyncIO dependency? Git's not idea for storing binaries. I think the process you're describing is git submodules, which are a possibility.

Are you experiencing a problem from the annotations dependency? It's not required at runtime -- only at compile time, as far as I know.

awb99:
If I have to recompile the solution a SubModule is much better; say I have to use a different library version; then with submodules this works much better.

In some other Issue I read that this is what someone asked to be done. I have compiled the solution with SharpDevelop and did not see any NuGet references. So I thought you have this somewhere in your build server. Without the JetBrain annotations library it would not compile.

drewnoakes:
It just sounds like your NuGet package restore is broken.

What version of NuGet are you using? I'm on 3.2.0.10516.

Are you using Mono?

tobi-tobsen:
I don't like the SubModule approach. Lets keep using nuget instead of git magic to pull unbuilt dependencies.

awb99:
I am using SharpDevelop on Windows; but I run the binaries on Mono. It seems SharpDeveop has some issues with NuGet. But the final dlls work perfect on Mono!! Amazing! Microsoft.Remoting Channels do not work on Mono for example.

somdoron:
great news. @awb99 is it ok if we close this PR and not merging, I think we pretty happy with nuget managing our dependencies and resharper annotations

awb99:
of course!

PR #449, opened 2016-01-25
Make it so NetMQSockets supply their own ZmqSocketType ids.

MikeFair:
Added a static property called TypeId on each Child Class that returns its ZmqSocketType.  Added an abstract readonly property called SocketType on NetMQSocket.  Then had the implementation of SocketType on each child class return its own class TypeId. Ultimately this is heading toward eliminating the switch statements based on the SocketType, move the procedures into the Socket classes directly, and provides a Socket constructor that takes a NetMQContext as a parameter.

### DIFF
--- a/src/NetMQ/NetMQSocket.cs
+++ b/src/NetMQ/NetMQSocket.cs
@@ -22,6 +22,14 @@ namespace NetMQ
         private EventHandler<NetMQSocketEventArgs> m_sendReady;
         private bool m_isClosed;
 
+        /// <summary>
+        /// The Socket Class type identifier for this Socket
+        /// </summary>
+        public abstract ZmqSocketType SocketType
+        {
+            get;
+        }
+
         internal enum DefaultAction
         {
             Connect,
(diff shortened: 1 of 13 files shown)


src/NetMQ/NetMQSocket.cs
```diff
@@ -22,6 +22,14 @@ public abstract class NetMQSocket : IOutgoingSocket, IReceivingSocket, ISocketPo
...
         private bool m_isClosed;
 
+        /// <summary>
+        /// The Socket Class type identifier for this Socket
+        /// </summary>
+        public abstract ZmqSocketType SocketType
+        {
+            get;
+        }
+
```
somdoron: You can get the socket type through SocketOptions.Type

MikeFair: Thanks, my oversight; didn't check well enough beforehand.  I'll refactor
it but likely will delete the reauest and wait until I have something a bit
more meaningful before trying again.

This request was trying to be a step towards isolating Sockets so they can
be inplemented using their own separate assembly modules without modifying
or recompiling the NetMQ module; same goal for transports too.

Thanks!

PR #451, opened 2016-01-25
problem: context might block forever is socket is not disposed or linger is not zero

### LINKED ISSUE #384: Feature Request: allow Context disposal even if it has sockets
Currently, if a context has created sockets, and they're doing things (or are even idle), attempts to Dispose() the context without disposing its sockets causes things to stall (see https://github.com/zeromq/netmq/issues/383 ). This is a minimal repro:

```
 class Program
    {
        static void Main(string[] args)
        {
            using (var client = new Client("tcp://127.0.0.1:7777"))
            {
                var cancel = new CancellationTokenSource();
                Task.Factory.StartNew(()=> client.Start(cancel.Token), TaskCreationOptions.LongRunning);
                Console.ReadLine();
                cancel.Cancel();
                //Process.GetCurrentProcess().Kill();
            }
        }
    }

    public class Client : IDisposable
    {
        private readonly string _endpoint;

        public Client(string endpoint)
        {
            _endpoint = endpoint;
            _ctx = NetMQContext.Create();
        }

        public void Start(CancellationToken token)
        {
            {
                _socket = _ctx.CreateRequestSocket();
                _socket.Options.Linger = TimeSpan.Zero;
                
                Thread.Sleep(1000);
                //_socket.Connect(_endpoint);

                while (!token.IsCancellationRequested)
                {
                    //_socket.SendMessage(new NetMQMessage(new[] { new NetMQFrame("This is my request") }));
                    Thread.Sleep(100);
                    //var msg = _socket.ReceiveMessage();
                    //Console.WriteLine("Received {0}", msg.First.ConvertToString());
                    Console.WriteLine("Received");
                }

            }
        }

        private bool _disposed;
        private readonly NetMQContext _ctx;
        private NetMQSocket _socket;

        public void Dispose()
        {
            if (_disposed) return;
            _disposed = true;
            _ctx.Dispose();
        }
    }
```
It would be good if there were an option to terminate the context - sockets and all - via Dispose (or some other method). As object disposal / finalizer threads can't be controlled in .NET, it'd remove a lot of ceremony around disposal.

somdoron:
solution: allow context to terminate without waiting for sockets

For NetMQContext you can use the Block keyword.
When creating sockets without context this is the default behavior, as the context is disposed when application exists.

fixes #384

### DIFF
--- a/src/NetMQ/Core/Ctx.cs
+++ b/src/NetMQ/Core/Ctx.cs
@@ -158,6 +158,11 @@ namespace NetMQ.Core
         private int m_ioThreadCount = DefaultIOThreads;
 
         /// <summary>
+        /// Should the context termination block until all socket are disposed?
+        /// </summary>
+        private bool m_block = true;
+
+        /// <summary>
         /// This object is used to synchronize access to context options.
         /// </summary>
         private readonly object m_optSync = new object();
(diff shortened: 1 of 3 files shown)


ashic:
Oh man...this just renders so many late night debugging sessions unnecessary. Awesome!!

Issue #424, opened 2016-01-10
Disposing socket and linger


somdoron:
To start, Socket.Dispose is actually async operation which continues in the background, when calling context.Dispose we actually wait for all the background operations to be complete.

Now that we removed the need for context the last socket get disposed is actually clearing the library, therefore it might get block as context get blocked if linger is not zero and still messages to send. This is true only if you are creating socket without context, if you still use the context you get old behavior.

The problem is that it is not intuitive, socket.dispose get blocked because of another socket?
In the past only the context.dispose was getting block.

take a look at following code

``` csharp
            DealerSocket p1 = new DealerSocket();
            DealerSocket p2 = new DealerSocket();
            DealerSocket justSocket = new DealerSocket();

            p2.Options.ReceiveHighWatermark = 10;
            p1.Options.SendHighWatermark = 10;

            p1.Bind("tcp://localhost:55555");
            p2.Connect("tcp://localhost:55555");

            int count = 0;

            p1.SendFrame("echo");

            // we fill the queue
            while (p1.TrySendFrame(TimeSpan.FromSeconds(1), "echo"))
            {
                count++;
            }

            Console.WriteLine("Blocked after {0}", count);

            p1.Dispose(); // this is causing the problem
            p2.Dispose();
            justSocket.Dispose(); // will get blocked
```

Just socket will get blocked, very hard to find.

I already have a fix that will make p1.Dispose to block. However this is breaking change as well. Users which didn't care about disposing the context never cared about this behavior. new version can break code and make dispose get blocked all around application.

Few solution:
1. Forget about cleaning the library (stopping the io thread) ,it is a background thread, just exit the app.
2. Users who care about clearing the library can call Global.Clean() or something.
3. Changing the linger default to zero or 1 second, also break change, however I think is subtle and most users don't know about.
4. When creating library register for AppDomain.CurrentDomain.ProcessExit and clean the library when process exit - the problem is that now someone won't understand why application won't exit, we can make a global property to disable this behavior.

I prefer the 4 solution.

@drewnoakes @ashic what do you think?

MikeFair:
I think something to also consider is how long it has been since the peer endpoints have sent/received any messages (can/is the physical connection up)?

[...]

PR #453, opened 2016-01-26
problem: dispose a socket is not blocking and when used without a context you have no way to know the status of the socket

somdoron:
solution: make the socket block on dispose.

This change is only for sockets created without socket, regular socket dispose will continue to be async.


somdoron:
Also added, last socket to get disposed close the global context.

somdoron:
fixes https://github.com/zeromq/netmq/issues/424

PR #455, opened 2016-01-28
Upgraded zproto for NetMQ.Zyre

dalebrubaker:
@somdoron , I think NetMQ.zproto is wonderful.
But, if I read this correctly, there is quite a bit of work to do on zproto_codec_cs.gsl to be able to generate a clean ZreMsg.cs class (in particular, to handle the msg type).
Yet it seems to me that this work should be first, rather than "rolling by hand" the ZreMsg.cs class.
What do you recommend? Should @vadian and/or I dig into changing zproto_codec_cs.gsl, or leave it to you and continue ZreMsg.cs without zproto?
Thank you.

### DIFF
--- a/src/NetMQ.Zyre.Tests/ZreMsgTests.cs
+++ b/src/NetMQ.Zyre.Tests/ZreMsgTests.cs
@@ -271,9 +271,12 @@ namespace NetMQ.Zyre.Tests
 
 			using (var client = new DealerSocket())
 			using (var server = new RouterSocket())
-			{
-				server.Bind("inproc://zprototest");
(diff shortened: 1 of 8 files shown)


dalebrubaker:
@somdoron ,
Um, on further examination I believe I mispoke. zproto_codec_cs.gsl maybe done and I didn't read ZerMsg.cs carefully enough. I'll continue to debug some more.
Sorry.

dalebrubaker:
Now the NetMQ.Zyre unit tests pass. ZerMsg.cs is done, first pass, subject to additional testing later.

PR #458, opened 2016-01-29
Additional work on NetMQ.Zyre

dalebrubaker:
In case @vadian or someone else is also working in this space.
ZrePeer is "done" first pass.
ZreGroup is barely started.


somdoron:
@dalebrubaker you have compilation errors...

PR #473, opened 2016-02-05
Remove NetMQ.Zyre projects; will move soon to separate repository

dalebrubaker:
Remove NetMQ.Zyre projects for now. Will move before too long to a separate repository, probably netmq/zyre

Also minor changes to NetMQBeacon that may be useful to Zyre


dalebrubaker:
Or, of course, just let me know if you prefer a different approach. :)

Issue #475, opened 2016-02-07
Problem: NetMQActor doesn't support variable arguments as CZMQ does


dalebrubaker:
czmq supports variable arguments that are passed to the shim.
see https://github.com/zeromq/czmq/blob/master/src/zactor.c where the signature is:
   zactor_t * zactor_new (zactor_fn *actor, void *args)

zyre uses this to pass both the shim (pipe for control of zre_node) and an outbox (another PairSocket used for a separate pipe of ZRE messages from peers back to the api)
see https://github.com/zeromq/zyre/blob/master/src/zyre_node.c
at the bottom is zyre_node_actor (zsock_t *pipe, void *args)

python also supports this variable-arguments signature in zeromq/pyre.
see https://github.com/zeromq/pyre/blob/797f01323089c64d08609db4dab387d777fa9b98/pyre/zactor.py
 
Adding params object[] args to the NetMQActor constructors seems not so trivial, because we would break existing users if we changed the IShimHandler to Run(PairSocket, params object[] )

But I don't currently know how to solve this problem, except to make NetMQActor accept both the existing IShimHandler and a new IShimHandlerArgs. This means (I think) duplicating all the ctors, the ShimAction delegates, and more. 

Before I make a mess, comments would be appreciated.

drewnoakes:
We are making breaking changes in the upcoming NetMQ4, so I would think
it's ok to make the change you want to IShimHandler so long as you don't
need this for the 3-x family of releases.

Ordinarily though, you're right, we can't be changing APIs without
backwards compatibility (as per C4.1 process).

somdoron:
You can just pass the parameter in the constructor of the class implementing IShimHandler.
This is how I do it.

dalebrubaker:
I thought so. But I'm stuck on the way zyre does this at https://github.com/zeromq/zyre/blob/master/src/zyre_node.c
 
At the bottom, you see that the zyre_node_actor (zsock_t *pipe, void *args) is creating the object so that it can loop and then eventually dispose/destroy the object. So there is no opportunity to pass in the extra outbox socket. If I create node in zyre where the actor is created, then I would somehow have to learn to dispose the node.

dalebrubaker:
Got it, thank you. Closing this issue.

PR #476, opened 2016-02-07
Breaking change: add params args to IShimHandler. Issue #475

dalebrubaker:


somdoron:
I don't think to make breaking change...

dalebrubaker:
Okay. This is what it would have looked like. Ref Issue #475 discussion.
So you can reject this PR...

dalebrubaker:
Actually, I think I see an easier way to make this not-breaking, so don't reject it yet. Thanks.

somdoron:
So my idea at the time was to just accept it as constructor parameter.
this is why I didn't not include it in the actor.
So you have to implement IShimHandler, in the class that implement just take a parameter. Save it in constructor and the Run method can use it.

dalebrubaker:
That works, thank you. Closing this PR.

PR #477, opened 2016-02-07
Added sample project that uses Reactive Extensions with NetMQ as the transport layer

sharpe5:
Attn: @drewnoakes @somdoron 

Just created a pull request that layers RX over NetMQ.

Using this wrapper over NetMQ, we can effortlessly use RX to send messages between different processes or machines.

The API is a 100% compatible, drop-in replacement for `Subject<T>` from Reactive Extensions.

As a refresher, to use `Subject<T>` in Reactive Extensions:

> Subject<int> subject = new Subject<int>();
	subject.Subscribe(message =>
	  {
		// Receives 42.
	  });
	subject.OnNext(42);

The new API is virtually identical:

> SubjectNetMQ<int> subject = new SubjectNetMQ<int>("tcp://127.0.0.1:56001");
	subject.Subscribe(message =>
	  {
		// Receives 42.
	  });
	subject.OnNext(42);

Currently, serialization is performed using ProtoBuf, so we have to annotate POCO classes with `ProtoContract` and `ProtoMember` like this:

> [ProtoContract]
	public struct MyMessage
	{
		[ProtoMember(1)]
		public int Num { get; set; }
		[ProtoMember(2)]
		public string Name { get; set; }
	}

To check out the demos, see:
- Publisher: Project `NetMQ.ReactiveExtensions.SamplePublisher`
- Subscriber: Project `NetMQ.ReactiveExtensions.SampleSubscriber`

Notes:
- 100% compatible, drop in replacement for `Subject<T>`.
- Allows messages to be passed across process and machine boundaries.
- Compatible with all existing RX code. Can use `.Where()`, `.Select()`, etc.
- Runs at 120,000 messages per second on localhost.
- Supports `.OnNext()`, `.OnException()`, and `.OnCompleted()`.
- Properly passes exceptions across the wire.


sharpe5:
Have to fix failing check; I think it's because I was testing compatibility with C# 5.0.

sharpe5:
Fixing build now...

PR #512, opened 2016-03-20
Problem: build.bat may fail due to NuGet references

opedroso:
Solution: Added NuGet command to download referenced packages prior to build
	  This requires user to download https://nuget.org/nuget.exe or
	  if preferred, build at least one using DevStudio NetMQ.sln, which
	  automatically downloads the referenced packages. After that first
	  build, build.bat will work with no errors even if NuGet.exe is not
	  present.

### DIFF
--- a/src/build.bat
+++ b/src/build.bat
@@ -1,5 +1,12 @@
 @ECHO OFF
 :: Usage:     build.bat [Clean]
+::
+::  Requires NuGet.exe to get references NuGets prior to command line build.
+::  If you prefer to not install NuGet.exe fom https://nuget.org/nuget.exe),
+::  build first time from NetMQ.sln using DevStudio, and the referenced NuGet
+:;  packages will be downloaded for you.
+::  After that first build, you can then use this script to build with no errors.
+::
 @setlocal
 
 set ACTION=Building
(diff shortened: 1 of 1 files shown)


src/build.bat
```diff
@@ -1,5 +1,12 @@
 @ECHO OFF
 :: Usage:     build.bat [Clean]
+::
+::  Requires NuGet.exe to get references NuGets prior to command line build.
+::  If you prefer to not install NuGet.exe fom https://nuget.org/nuget.exe),
+::  build first time from NetMQ.sln using DevStudio, and the referenced NuGet
+:;  packages will be downloaded for you.
```
c-rack: Small typo here?

opedroso:
Hi C-Rack,

Where would the typo be?
If you mean the extra parentheses in this line:

> +::  If you prefer to not install NuGet.exe fom https://nuget.org/nuget.exe),

then yes, the parentheses should not be there.

c-rack:
I meant the semicolon at the beginning of this line:
`:;  packages will be downloaded for you.`

opedroso:
Did not see it. You are correct, but it won't will break anything.
The scripts have been tested and it works fine.
 I will fix it though.

Thanks for noticing.

PR #519, opened 2016-03-30
Added Subscriber tests for Cleanup on Unbind

pcalin:


src/NetMQ/NetMQ.nuspec
```diff
@@ -16,9 +16,6 @@
     </releaseNotes>
     <copyright>Copyright 2016</copyright>
     <tags>ZeroMQ 0MQ CLRZMQ NetMQ Messaging ZMQ transport distributed</tags>
-    <dependencies>
-      <dependency id="AsyncIO" version="0.1.18.0" />
```
drewnoakes: Why did you remove AsyncIO?

src/NetMQ/packages.NetMQ.config
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="AsyncIO" version="0.1.18.0" targetFramework="net40" />
-  <package id="JetBrains.Annotations" version="10.0.0" targetFramework="net40" />
+  <package id="JetBrains.Annotations" version="10.0.0" targetFramework="net40" developmentDependency="true"/>
```
drewnoakes: Interesting and probably worthwhile though has no effect on our current way of packing for NuGet as we list dependencies in the nuspec file.

http://nuget.codeplex.com/workitem/1956

For completeness this might also be included in the .NET 3.5 .config file too.

src/NetMQ/NetMQ.nuspec
pcalin: I did not. I made sure nuget will properly use the config in the project. AsyncIO becomes a dependency because it is in the config.

src/NetMQ/packages.NetMQ.config
pcalin: See above. Because I removed the explicit dependencies from nuspec, nuget will now correctly bring in the real dependencies ignoring all developer dependencies.

I will add the same for .Net 3.5

pcalin:
Is there anything else holding this back, guys?

PR #522, opened 2016-03-30
Start using C# 6 language features

### LINKED ISSUE #520: Upgrade build to use Visual Studio 2014
It would be nice to be able to use C# 6 syntax.

drewnoakes:
Evolution of #366. Closes #520.

### DIFF
--- a/src/NetMQ/NetMQSocket.cs
+++ b/src/NetMQ/NetMQSocket.cs
@@ -159,15 +159,9 @@ namespace NetMQ
         /// <summary>
         /// Get the underlying <see cref="SocketBase"/>.
         /// </summary>
-        internal SocketBase SocketHandle
-        {
-            get { return m_socketHandle; }
-        }
+        internal SocketBase SocketHandle => m_socketHandle;
 
-        NetMQSocket ISocketPollable.Socket
-        {
-            get { return this; }
-        }
+        NetMQSocket ISocketPollable.Socket => this;
 
         #region Bind, Unbind, Connect, Disconnect, Close
 
@@ -412,10 +406,7 @@ namespace NetMQ
         /// Get whether a message is waiting to be picked up (<c>true</c> if there is, <c>false</c> if there is none).
         /// </summary>
         /// <exception cref="TerminatingException">The socket has been stopped.</exception>
-        public bool HasIn
-        {
-            get { return GetSocketOptionX<PollEvents>(ZmqSocketOption.Events).HasIn(); }
-        }
+        public bool HasIn => GetSocketOptionX<PollEvents>(ZmqSocketOption.Events).HasIn();
 
         /// <summary>
         /// Get whether a message is waiting to be sent.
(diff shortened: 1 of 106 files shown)


drewnoakes:
Seems TeamCity/Codebetter are not configured for C# 6 language features.

http://teamcity.codebetter.com/viewLog.html?buildId=214454&buildTypeId=bt1046&tab=buildLog

pcalin:
I think the build process is not using VS 14 hence cannot use C# 6. I had to roll back my C# 6 code last time I pushed.

drewnoakes:
Agreed. I think @somdoron needs to comment here.

somdoron:
We can try and use appvayor instead of codebetter, it might work. I will give it a try today

somdoron:
I managed to make it work on codebetter as well. No it is failing on one of the tests, I think it is a flapping test. Anyway I think it is ready to be merged...

src/NetMQ/NetMQSocket.cs
```diff
@@ -343,18 +343,10 @@ internal void InvokeEvents(object sender, PollEvents events)
             m_socketEventArgs.Init(events);
 
             if (events.HasIn())
-            {
-                var temp = m_receiveReady;
-                if (temp != null)
-                    temp(sender, m_socketEventArgs);
-            }
+                m_receiveReady?.Invoke(sender, m_socketEventArgs);
```
somdoron: I'm not sure how null propagation is working, but this might be not safe if null propagation is not copying the variable before the null check. the temp variable is used to avoid the race condition.

iiwaasnet:
appvayor will work for sure, I have a projects there in C# 6

src/NetMQ/NetMQSocket.cs
pcalin: Null propagation does a copy first. Identical to the code you have there. 


Sent from my Samsung Galaxy smartphone.-------- Original message --------From: Doron Somech <notifications@github.com> Date: 31/03/2016  18:25  (GMT+08:00) To: zeromq/netmq <netmq@noreply.github.com> Cc: Calin <pcalin@yahoo.com> Subject: Re: [zeromq/netmq] Start using C# 6 language features (#522) 
In src/NetMQ/NetMQSocket.cs:
> @@ -343,18 +343,10 @@ internal void InvokeEvents(object sender, PollEvents events)
>              m_socketEventArgs.Init(events);
>  
>              if (events.HasIn())
> -            {
> -                var temp = m_receiveReady;
> -                if (temp != null)
> -                    temp(sender, m_socketEventArgs);
> -            }
> +                m_receiveReady?.Invoke(sender, m_socketEventArgs);

I'm not sure how null propagation is working, but this might be not safe if null propagation is not copying the variable before the null check. the temp variable is used to avoid the race condition.

drewnoakes: Yep, this is a safe transformation thankfully. Makes the code read a lot more nicely.

drewnoakes:
That test seems to work most of the time, but does fail on the CI server occasionally: 

http://teamcity.codebetter.com/project.html?projectId=NetMQ&testNameId=1371961195580564611&tab=testDetails

I can't seem to make it fail locally though.

pcalin:
I can reproduce the problem. When a socket is disposed there is no telling when the reaper thread actually disposes the underlaying windows socket. So it is possible to attempt a bind on a socket that is still open.

I have created an extension method to help us in our code using a NetMqMonitor to wait until the socket is really disposed.

I'm surprised those tests don't fail more often.

pcalin:
It would be nice if there was a better/cleaner way to find out if socket is really disposed.

PR #535, opened 2016-04-26
Updated a few code samples in the documentation

yallie:
Updated a few code samples to match the current API.


drewnoakes:
Nice.

PR #536, opened 2016-04-27
Manual context termination

dgsantana:
This pull request, adds the ability to terminate and create the global NetMQ context. For this to happen a ManualTerminationTakeOver, must be called at least once. After this the developer is responsible for the life time control of the NetMQ context.
This is a requirement for Unity3D development where the Editor wasn't playing nice with the static global context, since after calling Play, the context isn't terminated (no application exit happens). This may also help in some extreme cases like using pythonnet with NetMQ inside a DCC app, like 3dsmax or maya, where there is need for more control over the life of the context.

### DIFF
--- a/src/NetMQ.Tests/ManualContextTerminationTests.cs
+++ b/src/NetMQ.Tests/ManualContextTerminationTests.cs
@@ -1,0 +1,145 @@
+﻿using System;
+using NetMQ.Sockets;
+using NUnit.Framework;
+
+namespace NetMQ.Tests
+{
+    [TestFixture]
+    public class ManualContextTerminationTests
+    {
+        [TestFixtureSetUp]
+        public void Setup()
+        {
+            NetMQConfig.ManualTerminationTakeOver();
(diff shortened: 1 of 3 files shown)


somdoron:
@dgsantana the CycleCreateTerminate test is hanging in the CI, can you check?

dgsantana:
Found the cue, its after running the BeaconTest that the
CycleCreateTerminate fails, so something must be left after running that
test. I'm going to debug the test and post results here.

Doron Somech <notifications@github.com> escreveu no dia segunda, 2/05/2016
às 16:16:

dgsantana:
Going a bit deeper, Beacon causes any blocking context termination to .
When waiting for the termination maibox to receive the Command.Done, the
signaler Wait never returns.

Daniel Santana <danielgsantana@gmail.com> escreveu no dia segunda,
2/05/2016 às 17:53:

somdoron:
So beacon... I will also give it a look

PR #543, opened 2016-05-05
Problem: Beacon didn't disposed the poller

somdoron:


dgsantana:
Awesome, @somdoron , I didn't had time during the week to do some proper testing/debugging.

d79ima:
i am still seeing this same exact exception when trying to call NetMQConfig.Cleanup() on application shut down. i am calling it right after poller.stop and beacon.silence, using netmq version 4.0.1.6

PR #580, opened 2016-07-11
Problem: Devices and Security only supported on .net4

somdoron:
Instead of fixing that I'm moving them to there own repository, that way the different packages will be almost equal.

new repositories are:
https://github.com/NetMQ/NetMQ.Devices
https://github.com/NetMQ/NetMQ.Security


drewnoakes:
Is this ready to be merged?

somdoron:
Yes

drewnoakes:
Sorry for the delay. Looks good.

PR #588, opened 2016-07-18
Fix tests on travis CI

somdoron:
please let the CIs runs.

Travis should build dontet core on Linux...


somdoron:
All is passing,  can be merged

PR #599, opened 2016-08-06
problem: Polling once with NetMQPoller is complicated

somdoron:
Expose internal selector class for advanced usages of polling


drewnoakes:
Very cool.

Issue #554, opened 2016-05-22
src/samples/weatherupdater: catastrophic error


mabra:
Hi All !

I am trying out 0MQ and the WeatherUpdater failes too.

> Console.WriteLine("Collecting updates for weather service for zipcode {0}...", zipToSubscribeTo);
> ::
> `
> Unhandled Exception: NetMQ.NetMQException: Exception of type 'NetMQ.NetMQException' was thrown.
>    at NetMQ.Core.Transports.Tcp.TcpConnector.OutCompleted(SocketError socketError, Int32 bytesTransferred)
>    at NetMQ.Core.Utils.Proactor.Loop()
>    at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
>    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
>    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
>    at System.Threading.ThreadHelper.ThreadStart()
> `

Inserting try/catch blocks does NOT catch the exception - the app dies !!
Spending an AppDomain.UnhandledException EventHandler shows "IsTerminating = true".

Changing port and/or address does'nt help much, but indeed, the exception disappears, if
I use "127.0.0.*". But no packets are on the wire (looking with windump) and the serverport
is not open.

I am now really disappointed.

The sample was run on:
Windows Server 2008 R2, en, all patches
Net 4.5

Regards,
Manfred

somdoron:
try to catch the socket error in TcpConnector line 234, for some reason the socket error is unknown and we need to figure out why.

MikkelPorse:
I have just seen this in my web application, SocketError was an OperationAborted (995), took the application pool down :o/

Without having thought long and hard about it, I would suggest adding it to the list of "recoverable" errorCodes ...

tobi-tobsen:
From reviewing the the patch #606 I think this can be closed now. @mabra and @MikkelPorse feel free to reopen it if it doesn't solve it.

PR #606, opened 2016-08-23
Added SocketError.OperationAborted to list of recoverable errors

MikkelPorse:
Possibly fixing issue 554.


drewnoakes:
Cross referencing #554

Issue #621, opened 2016-10-16
Missing NetMQPoller.ContainsSocket(NetmqSocket)


MartijnWoudstra:
I am trying to update my NetMQ. Since NetMQScheduler and Poller are obsolete, i am trying to use 
NetMQPoller. This method does not contain a ContainsSocket method to check if a socket is connected. 

The following code should fix it.

`public boolean ContainsSocket(NetMQSocket socket)
        {
            return m_sockets.Contains(socket)
        }
`

A pull request is also provided

drewnoakes:
To test whether a socket is contained within the poller, the `m_pollinSockets` must be queried on the poller thread. However any thread could call the `Contains` method.

One solution is to introduce `ContainsAsync` which would return a `Task<bool>`.

Making any `Contains` method block for the poller thread seems like opening the possibility for deadlocks.

PR #622, opened 2016-10-16
Update NetMQPoller.cs

MartijnWoudstra:
Update NETMQPoller, conforming issue #621


somdoron:
Take a look at the other methods in order to make it thread safe

drewnoakes:
I'd also vote for naming the method `Contains` (instead of `ContainsSocket`) so that it matches `Add` and `Remove`.

drewnoakes:
A thread safe implementation of this is in #650.

PR #634, opened 2017-01-09
Add traditional chinese translation for documents

liaochihung:
Hi,

I want to help the translation for the documents, but not sure in this way.

According to the http://docs.readthedocs.io/en/latest/localization.html#localization-of-documentation, it looks like I need to have a project there, then the main doc project https://netmq.readthedocs.io/en/latest/ can refer to other translated project, therefore I created an project in https://readthedocs.org/projects/netmq-traditional-chinese/.

Let me know if I done it in the wrong way, thanks.

### DIFF
--- a/docs/actor.md
+++ b/docs/actor.md
@@ -23,17 +23,17 @@ From [Wikipedia's Actor Model page](http://en.wikipedia.org/wiki/Actor_model):
 >
 > The Actor model is characterized by inherent concurrency of computation within and among actors, dynamic creation of actors, inclusion of actor addresses in messages, and interaction only through direct asynchronous message passing with no restriction on message arrival order.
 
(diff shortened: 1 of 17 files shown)


drewnoakes:
Hi,

Can you explain why you closed this? It'd be great to have your translation and I'm happy to help in any way I can.

There is already a French translation: http://netmq.readthedocs.io/fr/latest/

I think this must be configured by the owner of the readthedocs project. @somdoron is that you?

liaochihung:
Hi Drew Noakes,

Forgive my poor english...

I closed because it looks like not the right way, if you merge it, the
original doc will be replace with my translated one ?

And according to the
http://docs.readthedocs.io/en/latest/localization.html#localization-of-documentation
,
maybe the owner - somdoron, can add my translated
http://netmq-traditional-chinese.readthedocs.io/zh_TW/latest/
as child project of certain translation of zh_TW, therefore it can be the
http://netmq.readthedocs.io/zh_TW/latest/ ?

So the original NetMQ in github map to Readthedocs
http://netmq.readthedocs.io/en/latest/ as parent project
and my fork one map to Readthedocs
http://netmq-traditional-chinese.readthedocs.io/zh_TW/latest/ as a child
tranlation project.

btw, I've post the require in the mailing list too.

PR #655, opened 2017-04-08
Update version of AsyncIO to support by Windows XP (2003)

imiyu:


somdoron:
There is a conflict with the PR,  can you check?

Issue #663, opened 2017-04-24
Should NetMQPoller.Remove(Socket) throw if the socket is already disposed?


drewnoakes:
### Current behaviour

If a socket in the poller has been disposed, then `the `poller.Remove(ISocketPollable)` method will not throw. 

A disposed NetMQ socket in the poller can cause errors. `Select` operations internal to the poller may trigger `ObjectDisposedException`. It is not safe to have a disposed socket in the poller.

It is easy to get into that state, as discussed in #656. User code removes the socket from the poller, then disposes it:

```csharp
poller.Remove(socket);
socket.Dispose();
```

Problems related to this state may be quite infrequent, but may crash the program or log errors at exit.

Debugging these errors is hard because it is not obvious that `poller.Remove` can run asynchronously if not called from the poller thread, so it seems like a bug in NetMQ that the poller still contains the socket.

### Possible behaviour

We should make it easier to avoid getting into this state, as discussed in #656. However we can also help detect this state as soon as the bug is introduced.

Once detected, should we throw an exception? This would be the clearest way of getting a developer's attention, but it might start throwing exceptions in code that currently appears to work, because of the nature of the race condition. Personally I'd appreciate finding out, though not in production.

```csharp
public void Remove(ISocketPollable socket)
{
    if (socket == null)
        throw new ArgumentNullException(nameof(socket));
    if (socket.IsDisposed)
        throw new ArgumentException("Must not be disposed.", nameof(socket));
    CheckDisposed();

    Run(() =>
    {
        // Ensure the socket wasn't disposed while this code was waiting to be run on the poller thread
        if (socket.IsDisposed)
            throw new InvalidOperationException(
                $"{nameof(NetMQPoller)}.{nameof(Remove)} was called from a non-poller thread, " +
                "so ran asynchronously. " +
                $"The {socket.GetType().Name} socket being removed was disposed while the remove " +
                $"operation waited to start on the poller thread. Use {nameof(RemoveAndDispose)} " +
                "instead, which will enqueue the remove and dispose to happen on the poller thread.");

        socket.Socket.EventsChanged -= OnSocketEventsChanged;
        m_sockets.Remove(socket.Socket);
        m_isPollSetDirty = true;
    });
}
```

I can put a PR together for this, but want to hear whether introducing this new exception is a good idea or not.

It'd also require adding `ISocketPollable.IsDisposed`.

drewnoakes:
I edited the post to fix a few errors.

[...]

PR #664, opened 2017-04-24
Issue 656

drewnoakes:
- Adds `NetMQPoller.RemoveAndDispose(ISocketPollable)`
- Makes `NetMQPoller.Run` public (not in `net35)`

Also some general tidying and usage of C# 7 features such as local methods and inlined out var declarations.

This PR does not include any code for #663.


drewnoakes:
Just pushed another fix 5b1db5d to this PR. I think it's good for merge, but please review.

Issue #662, opened 2017-04-20
Many unit tests do not pass


Green7:
### Environment
    NetMQ Version:    
     last sources from github (20.0.4.2017)
    Operating System: 
     Windows 10, Visual Studio 2017
    .NET Version:
    4.5.1

### Actual behaviour

Many unit tests fail. 
Most of with "Cannot dispose poller from the poller thread." exception.
Some tests will never end. For example all test from class ProxyTests or InprocRouterDealerTest from 
SocketTests. It seems that they are hanging on NetMQConfig.Cleanup(); 

Is this normal behavior with current source code?

![capture](https://cloud.githubusercontent.com/assets/11303447/25228960/b0e2bfa2-25ce-11e7-93df-befc961547fd.PNG)

drewnoakes:
Some of those errors were fixed in #666.

Seems the CI systems (AppVeyor and TravisCI) aren't finding our tests since upgrading to VS2017's csproj format, so we're not seeing these bugs on PRs. I'm working on that in #665.

```text
dotnet test src\NetMQ.Tests\NetMQ.Tests.csproj --configuration Release --no-build
Test run for C:\projects\netmq-2bhss\src\NetMQ.Tests\bin\Release\net451\NetMQ.Tests.dll(.NETFramework,Version=v4.5.1)
Microsoft (R) Test Execution Command Line Tool Version 15.0.0.0
Copyright (c) Microsoft Corporation.  All rights reserved.
The path 'C:\projects\netmq-2bhss\src\NetMQ.Tests\bin\Release\net451' specified in the 'TestAdapterPath' does not contain any test adapters, provide a valid path and try again.
Starting test execution, please wait...
No test discoverer is registered to perform discovery of test cases. Register a test discoverer and try again.
Test run for C:\projects\netmq-2bhss\src\NetMQ.Tests\bin\Release\netcoreapp1.0\NetMQ.Tests.dll(.NETCoreApp,Version=v1.0)
Microsoft (R) Test Execution Command Line Tool Version 15.0.0.0
Copyright (c) Microsoft Corporation.  All rights reserved.
Starting test execution, please wait...
No test discoverer is registered to perform discovery of test cases. Register a test discoverer and try again.
```

We are using NUnit, which seems to be a bit behind on supporting newer .NET tooling and platforms. I'm investigating moving to xUnit which can handle these things and seems to be the more popular choice nowadays.

drewnoakes:
[...]

Issue #575, opened 2016-07-09
problem: NetMQPoller tests are failing on AppVeyor


somdoron:
For now I'm ignoring NetMQPoller test on AppVeyor

PR #665, opened 2017-04-24
Make unit tests run again on AppVeyor

drewnoakes:
Please don't merge this just yet.


drewnoakes:
I know how to make it build, but it seems that the tests aren't running. NUnit's not quite ready for .NET Core it seems. This may also explain #662. `dotnet test` in its various forms on my local machine also doesn't find/run the NUnit tests. Apparently they don't yet support VS2017's csproj format.

I've used xunit successfully on .NET core and AppVeyor. I'll investigate switching to xunit. Please shout if you see any problem with that.

somdoron:
I actually used custom libraries for NUnit because they didn't release a PR that I made. It is now released so you might want to try to use the official releases. It also what causing the problem, because I didn't update them recently.

somdoron:
Ok I see you already did it :-)

drewnoakes:
Oh I didn't realise that alternative lib was created by you!

I tried a few variations of getting NUnit to work. I'd prefer not having to change test framework, but it seems the NUnit stuff lags behind xUnit when it comes to support and community.

So far conversion to xUnit seems to be ok. By default xUnit runs tests in parallel, which doesn't seem to work with the global context.

somdoron:
I'm fine with changing the framework

drewnoakes:
I'll push the xUnit stuff to this PR for review shortly. It seems to work well, although occasionally some of the proxy tests throw `AddressAlreadyInUseException`, so there's something left to fix.

```text
NetMQ.AddressAlreadyInUseException : Cannot bind address ( inproc://frontend ) - already in use.
   at NetMQ.Core.SocketBase.Bind(String addr) in C:\Users\Drew\Dev\netmq\src\NetMQ\Core\SocketBase.cs:line 435
   at NetMQ.NetMQSocket.Bind(String address) in C:\Users\Drew\Dev\netmq\src\NetMQ\NetMQSocket.cs:line 183
   at NetMQ.Tests.ProxyTests.ControlSocketObservedMessages() in C:\Users\Drew\Dev\netmq\src\NetMQ.Tests\ProxyTests.cs:line 46
```

I'd appreciate some help working out why the proxy tests can fail.

drewnoakes:
Pushed. These tests pass on my machine, most of the time (proxy tests, as discussed).

src/NetMQ.Tests/ProxyTests.cs
```diff
@@ -1,29 +1,13 @@
...
+using Xunit;
 
 namespace NetMQ.Tests
 {
-    [TestFixture]
-    public class ProxyTests
+    public class ProxyTests : IClassFixture<NetMQConfigFixture>
     {
-        [OneTimeSetUp]
-        public void SetUp()
```
somdoron: I think this is the reason the tests fail.
Does xUnit has an alternative for this?

somdoron:
Try to call NetMQConfig.Cleanup before the proxy tests.

src/NetMQ.Tests/ProxyTests.cs
drewnoakes: The `IClassFixture<NetMQConfigFixture>` interface causes xUnit to run `NetMQConfigFixture.ctor` first, then run all the tests, then run `NetMQConfigFixture.Dispose`. So `Cleanup(true)` runs at the both the start and end of the fixture.

I'll try running it at the start and end of every test instead.

somdoron:
Let me know when to merge

drewnoakes:
I tried several combinations of settings for Travis to get a build working, and none of them succeeded. There are many pitfalls here. It seems that building a 2017 csproj that targets full .NET Framework on a Linux build server is not yet straightforward.

When I did find a combination of settings that got tests running in at least one configuration, they would hang and time out after ten minutes.

Despite this, I think this PR can now be merged. What it does do is get the unit tests working on AppVeyor. Currently the tests aren't running anywhere, so this is an improvement. If merged it should fix #662 and #575.

drewnoakes:
I think this PR is good to merge. It doesn't actually fix Travis CI, but it does make the tests run again on AppVeyor, as well as fix a few bugs that the tests would have picked up if they were running.

PR #670, opened 2017-05-06
Support for Universal Windows Platform apps

mqudsi:
Microsoft cheated with the System.Threading.Thread library: while it
is claimed that UWP implements .NET Standard 1.4 [0], and per the
listing at [1], System.Threading.Thread claims to support .NET
Standard 1.3, in reality the System.Threading.Thread package will
fail to install on UWP targets, despite them being fully .NET
Standard 1.3 complaint. If this were anyone else, the package would
probably not have been allowed to claim .NET Standard 1.3 support
in the nuspec listing.

As a result of this mess, for NetMQ to be usable under Universal
Windows Platform (i.e. Windows 10 and Windows 10 phone runtimes),
it is necessary to drop the dependency on System.Threading.Thread,
at least for this target. As a result, a new target has been added
to the project indepedendent from .NET Standard 1.3 so that all
other targets can continue using the System.Threading.Thread package.

The new UAP target for NetMQ has a fake implementation of threads
that uses the System.Threading.Tasks class internally to create a
Thread class sufficiently-resembling the real deal to compile and run
without issue. The nuspec for the NetMQ package might also need
updating to create a new target specifically for UAP consumers,
it would be identical to the .NET Standard 1.3 target except
without the System.Threading.Thread dependency.

[0]: https://github.com/dotnet/standard/blob/master/docs/versions.md
[1]: https://www.nuget.org/packages/System.Threading.Thread/

### DIFF
--- a/src/NetMQ/Core/Transports/Pgm/PgmSocket.cs
+++ b/src/NetMQ/Core/Transports/Pgm/PgmSocket.cs
@@ -98,7 +98,7 @@ namespace NetMQ.Core.Transports.Pgm
                 Debug.WriteLine(xMsg);
                 // If running on Microsoft Windows, suggest to the developer that he may need to install MSMQ in order to get PGM socket support.
 
-#if NETSTANDARD1_3
+#if NETSTANDARD1_3 || UAP
                 bool isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
 #else
                 PlatformID p = Environment.OSVersion.Platform;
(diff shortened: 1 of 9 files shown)


mqudsi:
If y'all prefer, instead of including `TaskThread.cs` directly, you can add a dependency to https://www.nuget.org/packages/TaskThread/ for the UAP target.

somdoron:
Thanks,  this is great

mqudsi:
Absolutely, glad to help. Thanks for this awesome project.

tomzorz:
Is there a time frame on the next release that has this in it? If I see correctly the 4.0.0.2-beta tag has this in it but it's not on NuGet.

somdoron:
I will try to push a version now...

tomzorz:
@somdoron Awesome, thank you!

somdoron:
pushed https://www.nuget.org/packages/NetMQ/4.0.0.4-beta

PR #700, opened 2017-09-05
Allow user to specify custom sync context

simonholt:
This is the sort of thing I have in mind.  This is backward compatible with any existing code, and shouldn't cause any behavioural changes to existing applications.


simonholt:
I'm seeing some strange behaviour in our unit tests - not sure whether this is the cause (I can't see how!) but I'm closing the PR until I get to the bottom of it...

simonholt:
#702 raised to cover the unit test problems I was having.

PR #702, opened 2017-09-10
Raise an error if removing a socket or timer from a stopped poller

simonholt:
I've been looking at a problem where my unit tests would lock up on a call to:

NetMQ.NetMQConfig.Cleanup(true);

The root cause is that my code was explicitly stopping the Poller before removing the sockets; therefore, the calls to Remove are never executed, and the Cleanup call waits forever.

I suggest raising an exception if this occurs, to make it more easy to detect this scenario.

### DIFF
--- a/src/NetMQ.Tests/NetMQPollerTest.cs
+++ b/src/NetMQ.Tests/NetMQPollerTest.cs
@@ -48,8 +48,6 @@ namespace NetMQ.Tests
 
                 Assert.Equal("World", req.ReceiveFrameString(out bool more2));
                 Assert.False(more2);
-
-                poller.Stop();
             }
         }
 
@@ -91,8 +89,6 @@ namespace NetMQ.Tests
                 Assert.True(listeningEvent.WaitOne(300));
                 Assert.True(connectedEvent.WaitOne(300));
                 Assert.True(acceptedEvent.WaitOne(300));
-
-                poller.Stop();
             }
         }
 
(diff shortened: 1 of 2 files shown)


drewnoakes:
Looks good. Definitely worth raising the error back to the user.

Wanted to add a note covering what I noticed during review.

There are some cases where the `Run(Action)` method that does the actual removal happens inline, so the exception could be skipped:

- on `NET35`, `Run(Action)` always runs inline
- on `!NET35`, `Run(Action)` runs inline if you're on the poller thread

I believe the second point is irrelevant as if `IsRunning` is false, but I'm not 100% sure that's the case.

It might be worth making the behaviour consistent between NET35 and !NET35. Anyone on NET35 that experiences this breaking change will be disappointed to know there's no real reason for it on their platform level, however folks upgrading from NET35 to !NET35, or multi-targetting, might find the difference in behaviour troublesome. On balance, I think we can merge as-is and discuss further if it turns out to be a problem (hence these notes).

Is there a way that `NetMQ.NetMQConfig.Cleanup(true)` could detect the deadlock? What were the relevant circumstances during the deadlock?

somdoron:
this actually causing me a problem, how this is helps with cleanup?

I have class in a library that accept a poller and register to it, when class disposed its removing itself from the poller, not caring if the poller is dead or alive.

Also I think it is valid to remove socket from not running poller (but not from disposed one), as you might even remove the socket before running the poller yet.

simonholt:
It looks like commit a6610a060f39a77915dcada6d4fc90b9ef90bc82 may be intended to resolve the same problem I had, in which case this isn't needed any more.  I'll try and repro the original problem and verify whether or not it's fixed at some point.

PR #708, opened 2017-11-02
problem: developing peer to peer with current socket is hard

somdoron:
you have to use a dealer socket for every connection you made.

With peer socket type, you can get routing id of the peer you just connected to and use one socket both for incoming connections and outgoing connections.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,7 +6,7 @@ branches:
 image: Visual Studio 2017
 configuration: Release
 before_build:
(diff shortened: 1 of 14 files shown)


somdoron:
@drewnoakes can you merge?

somdoron:
@c-rack can you please merge?

c-rack:
Sure :)

somdoron:
thanks

PR #711, opened 2017-11-06
Minor changes but vital for managing sudden network change…

IEliminator:
Minor changes but vital for managing network change in graceful manner. Sudden network change can result in crash if no try/catch when trying to send udp message or no check for array element swapping.

When changing network during execution (runtime) application crashes due to either SocketExcpetion or OutOfRange for swapping elements. Both cases only occurs during sudden network change, (e.g. going from a subnet 192.168.131 to 192.168.132 with 24bit subnet mask). Adding these simple checks will prevent from crashing and "on next iteration" wokring proparly. 

Please not that I have not added the code for creating new Udp  after the if statement in the catch block. I am not familiar with NetMQ internals and as such would like to leave it to you to consider this.

### DIFF
--- a/src/NetMQ/Core/Patterns/Utils/ArrayExtensions.cs
+++ b/src/NetMQ/Core/Patterns/Utils/ArrayExtensions.cs
@@ -45,7 +45,8 @@ namespace NetMQ.Core.Patterns.Utils
 
         public static void Swap<T>([NotNull] this List<T> items, int index1, int index2) where T : class
         {
-            if (index1 == index2) { return; }
+            if (index1 == index2) 
+                return;
 
             T item1 = items[index1];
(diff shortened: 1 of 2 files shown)


somdoron:
@IEliminator please remove the array extension fix as it is not correct and I will merge it.

IEliminator:
The ArrayExtension fix is removed but I don't understand really why there shouldn't be a validation of the list size and indexex that you want to act on. This is most likely due to my lack of understanding of the details behind the scenen so if you have the time I would appreciate minor explanation (for my understanding).
As noticed in execution environment when sudden (subnet) network switch occurs unhandle exception is thorwn due to ArgumentOutOfRange in this specific method. The simple condition control has proven to solve it...

PR #719, opened 2017-12-22
Issue716

mikepmiller:
Error handling for receive call on PGM socket.  See the note at WSAECONNRESET at WSARecv function: For a UDP datagram socket, this error would indicate that a previous send operation resulted in an ICMP "Port Unreachable" message.


drewnoakes:
Looks good to me. @somdoron do you see any issue with this?

mikepmiller:
@somdoron, saw that this was merged. Just wanted to say thanks, and more importantly, thanks for providing a fantastic library! I’ve used it for more than a year, and it’s been awesome.

Issue #644, opened 2017-03-09
Crash on Mono with System.Net.Sockets.SocketException (0x80004005)


redmanmale:
I've tried to move one of my projects to Mono and I was faced with the fact that it crashed on start.

I don't know where the problem is, in NetMq, AsyncIO or Mono, if I was mistaken to open issue here, please give me the cue where to move it.

### Environment
```
NetMQ Version: 4.0.0.1
AsyncIO Version: 0.1.26.0
Operating System: Win 7 x64
.NET Version: - 

>mono --version
Mono JIT compiler version 4.8.0 (Visual Studio built mono)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
        TLS:           normal
        SIGSEGV:       normal
        Notification:  Thread + polling
        Architecture:  amd64
        Disabled:      none
        Misc:          softdebug
        LLVM:          supported, not enabled.
        GC:            sgen
```

### Expected behaviour

Don't crash.

### Actual behaviour

Crash.
```
Unhandled Exception:
System.Net.Sockets.SocketException (0x80004005): Р?РїРчС?Р°С┼РёС? С?С?РїРчС?Р?Р? Р·Р°Р?РчС?С?РчР?Р°.


  at AsyncIO.Windows.Socket.SetSocketOption (System.Net.Sockets.SocketOptionLevel optionLevel, Syste
m.Net.Sockets.SocketOptionName optionName, System.Int32 optionValue) [0x00018] in <76dbd955e71949f7b
13045772e552e7a>:0
  at AsyncIO.AsyncSocket.set_NoDelay (System.Boolean value) [0x00000] in <76dbd955e71949f7b13045772e
552e7a>:0
  at NetMQ.Core.Transports.Tcp.TcpConnector.OutCompleted (System.Net.Sockets.SocketError socketError
, System.Int32 bytesTransferred) [0x00099] in <cd775f2ee43b4afd8cc83c9f5ddee1f4>:0
  at NetMQ.Core.Transports.Tcp.TcpConnector.StartConnecting () [0x00097] Message sent
in <cd775f2ee43b4afd8cc83c9f5ddee1f4>:0
  at NetMQ.Core.Transports.Tcp.TcpConnector.ProcessPlug () [0x0001b] in <cd775f2ee43b4afd8cc83c9f5dd
ee1f4>:0
[...]

PR #763, opened 2018-11-05
Only set AsyncSocket.NoDelay when not already set. Connected to #644

anders9ustafsson:
Connected to Github issue #644, see comment [here](https://github.com/zeromq/netmq/issues/644#issuecomment-435871734).

Appears to be a necessary precondition for building NetMQ with the .NET native toolchain for UWP applications. Have not verified that the fix applies to the platforms originally mentioned in the issue report.

All unit tests succeed after updating the source code.


somdoron:
thanks

PR #775, opened 2018-11-25
Fixed #774, Update attributes in function Msg.Put

### LINKED ISSUE #774: How about changing attributes, [CanBeNull] in src/NetMQ/Msg.cs
In function [Buffer.CopyBlock](https://docs.microsoft.com/en-us/dotnet/api/system.buffer.blockcopy?view=netframework-4.7.2), if src or dst is null, then it raises ArgumentNullException. So I think this shouldn't be null. But in Msg.cs source, function Put() allow null by using [CanBeNull] Attribute. So I suggest how about using attribute [NotNull] not [CanBeNull]

moreal:
CanBeNull to NotNull


somdoron:
Thanks

Issue #783, opened 2019-03-13
all source code wrong &gt;  <- this is error


avap90:
### Environment

    NetMQ Version:    v4.0.0.1
    Operating System: Windows 10
    .NET Version:     4.7.2

   site : https://netmq.readthedocs.io/en/latest/request-response/
  
### Expected behaviour
all source code wrong &gt ; <- this is error html source is <  but 
web page source code &gt ;

if user source code copy and past then runtime error &gt ; tcp://..... so error

avap90:
https://ibb.co/3CbdTcj

avap90:
[https://ibb.co/3CbdTcj](url)

moreal:
actually &gt; is same with '>'

In my thought, it is occured by `readthedocs`, they will filter '<' and '>' for restrict xss

somdoron:
can you make a pull request?

moreal:
it seems i found the way to solve

can I be assigned on this issue?

moreal:
https://netmq-forked.readthedocs.io/en/latest/request-response/

like this

avap90:
> can you make a pull request?

...... can i have permission?

avap90:
> https://netmq-forked.readthedocs.io/en/latest/request-response/
> 
> like this

nice!~~

moreal:
@avap90 @somdoron 

I made pull request, can you check the new docs?

avap90:
> @avap90 @somdoron
> 
> I made pull request, can you check the new docs?

Yes. I am checked.
But, not changed.... "&gt" same...

moreal:
@avap90 

The original page will be changed when the pr was accepted

moreal:
In my thought, it seems resolved by #785. Is there remained parts, yet?

PR #785, opened 2019-03-19
Update docs to avoid html-special-characters filtering

moreal:
I resolved issue #783 

In my thought, this problem was occured from using syntax not standard  
(The docs build program might think < and > can be used for xss attack.)

So I changed the code block format from ":::<lang>" => "``` <lang>``` using [code](https://gist.github.com/moreal/56471e375729e1f0693b08b52cd7adbd) (gist)

Please check it's successfully changed by me, [link, netmq-forked readthedocs](https://netmq-forked.readthedocs.io)


somdoron:
thanks!

PR #802, opened 2019-07-07
Problem: NetMQ cannot be used with async/await pattern

somdoron:
Solution: Add ReceiveAsync (and some overloads) to socket. Also, a NetMQRuntime must be created and run in order to signal the completion of the tasks

### DIFF
--- a/src/NetMQ/AsyncReceiveExtensions.cs
+++ b/src/NetMQ/AsyncReceiveExtensions.cs
@@ -1,0 +1,219 @@
+﻿#if NETSTANDARD2_0 || NET47
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using System.Diagnostics.CodeAnalysis;
+using System.Text;
+using JetBrains.Annotations;
+
+
+namespace NetMQ
+{
+    /// <summary>
+    /// Provides extension methods for the <see cref="NetMQSocket"/>,
+    /// via which messages may be received asynchronously.
+    /// </summary>
+    [SuppressMessage("ReSharper", "MemberCanBePrivate.Global")]
+    [SuppressMessage("ReSharper", "UnusedMember.Global")]
+    [SuppressMessage("ReSharper", "UnusedMethodReturnValue.Global")]
+    public static class AsyncReceiveExtensions
+    {
+        #region Receiving a frame as a byte array
+
+        /// <summary>
+        /// Receive a single frame from <paramref name="socket"/>, asynchronously.
+        /// </summary>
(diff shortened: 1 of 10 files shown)


src/NetMQ/AsyncReceiveExtensions.cs
```diff
@@ -0,0 +1,220 @@
...
+            return socket.ReceiveFrameStringAsync(SendReceiveConstants.DefaultEncoding);
+        }
+
+
+        /// <summary>
+        /// Receive a single frame from <paramref name="socket"/>, asynchronously, and decode as a string using <paramref name="encoding"/>.
+        /// </summary>
+        /// <param name="socket">The socket to receive from.</param>
+        /// <param name="encoding">The encoding used to convert the frame's data to a string.</param>
+        /// <returns>The content of the received message frame as a string and a  .</returns>
```
drewnoakes: Incomplete comment?

src/NetMQ/AsyncReceiveExtensions.cs
```diff
@@ -0,0 +1,220 @@
...
+    [SuppressMessage("ReSharper", "UnusedMethodReturnValue.Global")]
+    public static class AsyncReceiveExtensions
+    {
+        #region Receiving a frame as a byte array
+
+        /// <summary>
+        /// Receive a single frame from <paramref name="socket"/>, asynchronously.
+        /// </summary>
+        /// <param name="socket">The socket to receive from.</param>
+        /// <param name="more"><c>true</c> , otherwise <c>false</c>.</param>
```
drewnoakes: This overload doesn't have a `more` parameter.

src/NetMQ/AsyncReceiveExtensions.cs
somdoron: yes, fixing

somdoron:
When core 3 is out we can also implement IAsyncEnumerable

src/NetMQ/AsyncReceiveExtensions.cs
```diff
@@ -0,0 +1,219 @@
...
+            socket.AttachToRuntime();
+
+            var msg = new Msg();
+            msg.InitEmpty();
+
+            if (socket.TryReceive(ref msg, TimeSpan.Zero))
+            {
+                bool more = msg.HasMore;
+                msg.Close();
+                return Task.FromResult(more);
```
drewnoakes: Should probably cache a true result and a false result here to avoid allocating.

somdoron: addressed in https://github.com/zeromq/netmq/pull/804

PR #808, opened 2019-07-20
Fix problem where Cleanup hangs on macOS .NET Core

### LINKED ISSUE #806: NetMQConfig.Cleanup(false) hangs
### Environment

    NetMQ Version: 4.0.0.239-pre
    Operating System: macOS
    .NET Version:  .NETCore 2.2.203

### Expected behaviour

`NetMQConfig.Cleanup(false)` get stuck when it called.

### Actual behaviour

`NetMQConfig.Cleanup(false)` doesn't block current thread.

```
public class SwarmTest : IDisposable
{
    // some test cases with NetMQSocket

     public void Dispose()
     {
        NetMQConfig.Cleanup(false);
     }
}
```

earlbread:
This fixes #806.

Currently, `Socket.Select` does not work properly on macOS .NET Core when `readList` and `errorList` are passed together. To avoid this problem, I changed to call the `Select` function separately for `errorList`.

Please refer to this issue: https://github.com/dotnet/corefx/issues/39617

### DIFF
--- a/src/NetMQ/Core/Utils/Poller.cs
+++ b/src/NetMQ/Core/Utils/Poller.cs
@@ -23,6 +23,9 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net.Sockets;
+#if !NETFRAMEWORK
+using System.Runtime.InteropServices;
+#endif
 using System.Threading;
 using JetBrains.Annotations;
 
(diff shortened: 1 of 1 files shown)


somdoron:
Can you check the OS and pick the right strategy accordingly?

earlbread:
Fixed workaround to work only on macOS .NET core. @somdoron

somdoron:
Thanks!

PR #812, opened 2019-08-30
Added details about dispose ordering

Thieum:
I had to do some testing to find the right syntax, when instances are held instead of used in a using like the examples.


somdoron:
Thanks

Issue #823, opened 2019-10-08
Concurrency issues in Clock.cs


acizmarik:
Current implementation of `NowMs` is source of multiple concurrency issues which causes the method to return incorrect time. The issue is caused by usage of two static fields `s_lastTsc` and `s_lastTime`

https://github.com/zeromq/netmq/blob/057532db183ea001548a54a22ffebd630801c000/src/NetMQ/Core/Utils/Clock.cs#L88-L105

The condition `if (tsc - s_lastTsc <= Config.ClockPrecision / 2 && tsc >= s_lastTsc)` is in race with the subsequent assignments (both can be issued from different threads):
`s_lastTsc = tsc;`
`s_lastTime = NowUs() / 1000;`

Additionally when reading/writing 64bit fields, one must use `Interlocked.Read` or `Interlocked.Write` to ensure that these operations are atomic also for 32bit programs.

somdoron:
Want to make a pull request?

Andy1221:
Yes I will make one tomorrow or later today.

somdoron:
Thanks

PR #824, opened 2019-10-09
Fixing concurrency issues in Clock

acizmarik:
This is one of the possible solutions for issue: #823.
Alternatively both fields can be marked as ThreadStatic and original code would work.

I was not able to run performance tests locally - for me it crashes after a while saying something about port being already used. Therefore I am not 100% sure which approach would be better for performance.

### DIFF
--- a/src/NetMQ/Core/Utils/Clock.cs
+++ b/src/NetMQ/Core/Utils/Clock.cs
@@ -20,6 +20,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Threading;
 
 namespace NetMQ.Core.Utils
 {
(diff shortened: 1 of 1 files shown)


src/NetMQ/Core/Utils/Clock.cs
```diff
@@ -67,6 +70,7 @@ static Clock()
                 s_rdtscSupported = false;
             }
 #endif
+            s_spinLock = new SpinLock();
```
drewnoakes:
```suggestion
            s_spinLock = new SpinLock(enableThreadOwnerTracking: false);
```

I don't think we need thread ownership tracking ([see](https://docs.microsoft.com/en-us/dotnet/api/system.threading.spinlock.-ctor?view=netframework-4.8#remarks)).

PR #835, opened 2019-12-08
Poller.remove socket disposed fix

### LINKED ISSUE #834: Poller will throw ObjectDisposed exception if Socket.Dispose() called
### Environment

    NetMQ Version:    
    Operating System: 
    .NET Version:

jasells:
This should fix #834 and a few other similar potential bugs in other versions of ```Remove()```, and ```RemoveAndDispose()```

See a few other comments on a couple of the tests and about throwing ```ArgumentException``` from ```Remove()``` when socket is disposed.  Those should be left to a future PR, though.

### DIFF
--- a/src/NetMQ.Tests/NetMQPollerTest.cs
+++ b/src/NetMQ.Tests/NetMQPollerTest.cs
@@ -455,7 +455,7 @@ namespace NetMQ.Tests
             var pubThread = Task.Run(pubAction);
 
             //allow a little time to run
-            await Task.Delay(5000);
+            await Task.Delay(2000);
 
             //now try to remove the sub from poller
             await patient.Remove(sub);
@@ -465,7 +465,7 @@ namespace NetMQ.Tests
             sub = null;
 
             //allow for poller to continue running
-            await Task.Delay(3000);
+            await Task.Delay(2000);
 
             patient.Stop();
             Assert.False(patient.IsRunning);
--- a/src/NetMQ.Tests/NetMQPollerTest.cs
+++ b/src/NetMQ.Tests/NetMQPollerTest.cs
@@ -401,7 +401,7 @@ namespace NetMQ.Tests
         }
 
         [Fact]
-        public async void RemoveAndDisposeSocket()
+        public async void DisposeSocketAfterRemoval()
         {
             //set up poller, start it
             var patient = new NetMQPoller();
--- a/src/NetMQ.Tests/NetMQPollerTest.cs
+++ b/src/NetMQ.Tests/NetMQPollerTest.cs
@@ -458,10 +458,11 @@ namespace NetMQ.Tests
             await Task.Delay(5000);
 
             //now try to remove the sub from poller
-            patient.Remove(sub);
+            await patient.Remove(sub);
 
             // dispose the sub (this will cause exception on poller's worker-thread) and it can't be caught!
             sub.Dispose();
+            sub = null;
 
             //allow for poller to continue running
             await Task.Delay(3000);
@@ -471,9 +472,9 @@ namespace NetMQ.Tests
 
             canceller.Cancel();
 
-            //patient?.Dispose();
             pub?.Dispose();
-            sub?.Dispose();
+            sub?.Dispose();// left here for dev testing when prior .Dispose() call may be commented out
+            patient?.Dispose();
         }
 
         [Fact]
@@ -499,7 +500,7 @@ namespace NetMQ.Tests
         }
 
         [Fact]
-        public void RemoveThrowsIfSocketAlreadyDisposed()
+        public async void RemoveThrowsIfSocketAlreadyDisposed()
         {
             var socket = new RouterSocket();
 
@@ -511,7 +512,7 @@ namespace NetMQ.Tests
             socket.Dispose();
 
             // Remove throws if the removed socket
-            var ex = Assert.Throws<ArgumentException>(() => poller.Remove(socket));
+            var ex = await Assert.ThrowsAsync<ArgumentException>(() => poller.Remove(socket));
 
             Assert.StartsWith("Must not be disposed.", ex.Message);
             Assert.Equal("socket", ex.ParamName);
--- a/src/NetMQ.Tests/NetMQPollerTest.cs
+++ b/src/NetMQ.Tests/NetMQPollerTest.cs
@@ -583,19 +583,8 @@ namespace NetMQ.Tests
             socket.Dispose();
 
             // Remove throws if the removed socket
-            //var ex = Assert.Throws<ArgumentException>(() => poller.Remove(socket));
+            var ex = Assert.Throws<ArgumentException>(() => poller.Remove(socket));
 
-            ArgumentException ex = null;
-
-            try
-            {
-                poller.Remove(socket);
-            }
-            catch (ArgumentException argEx)
-            {
-                ex = argEx;
-            }
-            Assert.NotNull(ex);
             Assert.StartsWith("Must not be disposed.", ex.Message);
             Assert.Equal("socket", ex.ParamName);
 
--- a/src/NetMQ.Tests/NetMQPollerTest.cs
+++ b/src/NetMQ.Tests/NetMQPollerTest.cs
@@ -401,6 +401,82 @@ namespace NetMQ.Tests
         }
 
         [Fact]
+        public async void RemoveAndDisposeSocket()
+        {
+            //set up poller, start it
+            var patient = new NetMQPoller();
+            patient.RunAsync();
+            
+            Assert.True(patient.IsRunning);
+
+            //create a pub-sub pair
+            var port = 55667;
+            var conn = $"tcp://127.0.0.1:{port}";
+
+            var pub = new PublisherSocket();
+            pub.Bind(conn);
+
+            var sub = new SubscriberSocket();
+            sub.Connect(conn);
+            sub.SubscribeToAnyTopic();
+
+            //handle callbacks from poller thread
+            sub.ReceiveReady += (s, e) =>
+            {
+                var msg = e.Socket.ReceiveFrameString();
+
+                Debug.WriteLine($"sub has data: {msg}");
+            };
+
+            //add the subscriber socket to poller
+            patient.Add(sub);
+
+            //set up pub on separate thread
+            var canceller = new CancellationTokenSource();
+
+            var pubAction = new Action(async () =>
+            {
+                var token = canceller.Token;
+
+                uint i = 0;
+
+                while(!token.IsCancellationRequested)
+                {
+                    pub.SendFrame($"Hello-{++i}");
+
+                    // send ~ 5Hz
+                    await Task.Delay(200);
+                }
+            });
+
+            //var pubThread = new Task(pubAction, canceller.Token, TaskCreationOptions.LongRunning | TaskCreationOptions.AttachedToParent);
+            //pubThread.Start();
+
+            var pubThread = Task.Run(pubAction);
+
+            //allow a little time to run
+            await Task.Delay(5000);
+
+            //now try to remove the sub from poller
+            patient.Remove(sub);
+
+            // dispose the sub (this will cause exception on poller's worker-thread) and it can't be caught!
+            sub.Dispose();
+
+            //allow for poller to continue running
+            await Task.Delay(3000);
+
+            patient.Stop();
+            Assert.False(patient.IsRunning);
+
+            canceller.Cancel();
+
+            //patient?.Dispose();
+            pub?.Dispose();
+            sub?.Dispose();
+        }
+
+        [Fact]
         public void AddThrowsIfSocketAlreadyDisposed()
         {
             var poller = new NetMQPoller();
--- a/src/NetMQ.Tests/NetMQPollerTest.cs
+++ b/src/NetMQ.Tests/NetMQPollerTest.cs
@@ -347,6 +347,9 @@ namespace NetMQ.Tests
                     // identity
                     e.Socket.SkipFrame();
 
+
+                    //**Note: bad to assert from worker thread!  
+                    // If it fails, the test will crash, not report failure!
                     Assert.Equal("Hello", e.Socket.ReceiveFrameString(out bool more));
                     Assert.False(more);
 
@@ -401,6 +404,77 @@ namespace NetMQ.Tests
         }
 
         [Fact]
+        public async void RemoveAndDisposeSocket()
+        {
+            //set up poller, start it
+            var patient = new NetMQPoller();
+            patient.RunAsync();
+
+            Assert.True(patient.IsRunning);
+
+            //create a pub-sub pair
+            var port = 55667;
+            var conn = $"tcp://127.0.0.1:{port}";
+
+            var pub = new PublisherSocket();
+            pub.Bind(conn);
+
+            var sub = new SubscriberSocket();
+            sub.Connect(conn);
+            sub.SubscribeToAnyTopic();
+
+            //handle callbacks from poller thread
+            sub.ReceiveReady += (s, e) =>
+            {
+                var msg = e.Socket.ReceiveFrameString();
+
+                Debug.WriteLine($"sub has data: {msg}");
+            };
+
+            //add the subscriber socket to poller
+            patient.Add(sub);
+
+            //set up pub on separate thread
+            var canceller = new CancellationTokenSource();
+
+            var pubAction = new Action(async () =>
+            {
+                var token = canceller.Token;
+
+                uint i = 0;
+
+                while (!token.IsCancellationRequested)
+                {
+                    pub.SendFrame($"Hello-{++i}");
+
+                    // send ~ 5Hz
+                    await Task.Delay(200);
+                }
+            });
+
+            var pubThread = Task.Run(pubAction);
+
+            //allow a little time to run
+            await Task.Delay(2000);
+
+            //now try to remove the sub from poller
+            patient.RemoveAndDispose(sub);
+
+            Assert.True(sub.IsDisposed);
+
+            //allow for poller to continue running
+            await Task.Delay(2000);
+
+            patient.Stop();
+            Assert.False(patient.IsRunning);
+
+            canceller.Cancel();
+
+            pub?.Dispose();
+            patient?.Dispose();
+        }
+
+        [Fact]
         public async void DisposeSocketAfterRemoval()
         {
             //set up poller, start it
@@ -448,9 +522,6 @@ namespace NetMQ.Tests
                     await Task.Delay(200);
                 }
             });
-
-            //var pubThread = new Task(pubAction, canceller.Token, TaskCreationOptions.LongRunning | TaskCreationOptions.AttachedToParent);
-            //pubThread.Start();
 
             var pubThread = Task.Run(pubAction);
 
@@ -542,6 +613,11 @@ namespace NetMQ.Tests
             // Dispose the socket.
             // It is incorrect to have a disposed socket in a poller.
             // Disposed sockets can throw into the poller's thread.
+
+            //**JASells: And what does that have to do with removing one?  Should check for disposed 
+            // socket on Add, not Remove!  Check only internally on to avoid accessing a potentially,
+            // disposed socket, but otherwise, removing it from the list is safe, and preferable to 
+            // throwing exception since it makes the poller more resilient to unintended mis-use!
             socket.Dispose();
 
             // Dispose throws if a polled socket is disposed
--- a/src/NetMQ.Tests/NetMQPollerTest.cs
+++ b/src/NetMQ.Tests/NetMQPollerTest.cs
@@ -458,7 +458,7 @@ namespace NetMQ.Tests
             await Task.Delay(2000);
 
             //now try to remove the sub from poller
-            await patient.Remove(sub);
+            patient.Remove(sub);
 
             // dispose the sub (this will cause exception on poller's worker-thread) and it can't be caught!
             sub.Dispose();
@@ -500,7 +500,7 @@ namespace NetMQ.Tests
         }
 
         [Fact]
-        public async void RemoveThrowsIfSocketAlreadyDisposed()
+        public void RemoveThrowsIfSocketAlreadyDisposed()
         {
             var socket = new RouterSocket();
 
@@ -512,8 +512,19 @@ namespace NetMQ.Tests
             socket.Dispose();
 
             // Remove throws if the removed socket
-            var ex = await Assert.ThrowsAsync<ArgumentException>(() => poller.Remove(socket));
+            //var ex = Assert.Throws<ArgumentException>(() => poller.Remove(socket));
 
+            ArgumentException ex = null;
+
+            try
+            {
+                poller.Remove(socket);
+            }
+            catch (ArgumentException argEx)
+            {
+                ex = argEx;
+            }
+            Assert.NotNull(ex);
             Assert.StartsWith("Must not be disposed.", ex.Message);
             Assert.Equal("socket", ex.ParamName);
 
--- a/src/NetMQ/ISocketPollableCollection.cs
+++ b/src/NetMQ/ISocketPollableCollection.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading.Tasks;
 using JetBrains.Annotations;
 using NetMQ.Monitoring;
 
@@ -13,7 +14,7 @@ namespace NetMQ
     public interface ISocketPollableCollection
     {
         void Add([NotNull] ISocketPollable socket);
-        void Remove([NotNull] ISocketPollable socket);
+        Task Remove([NotNull] ISocketPollable socket);
         void RemoveAndDispose<T>(T socket) where T : ISocketPollable, IDisposable;
     }
 }
--- a/src/NetMQ/ISocketPollableCollection.cs
+++ b/src/NetMQ/ISocketPollableCollection.cs
@@ -14,7 +14,7 @@ namespace NetMQ
     public interface ISocketPollableCollection
     {
         void Add([NotNull] ISocketPollable socket);
-        Task Remove([NotNull] ISocketPollable socket);
+        void Remove([NotNull] ISocketPollable socket);
         void RemoveAndDispose<T>(T socket) where T : ISocketPollable, IDisposable;
     }
 }
--- a/src/NetMQ/NetMQPoller.cs
+++ b/src/NetMQ/NetMQPoller.cs
@@ -115,24 +115,37 @@ namespace NetMQ
 
         public Task Run([NotNull] Action action)
         {
-            Task t = new Task(action);
+            Task t;
 
             if (!IsRunning || CanExecuteTaskInline)
             {
                 action();
 
-                var tcs = new TaskCompletionSource<object>();
-                tcs.SetResult(null);
-
-                t = tcs.Task;
+                t = FromResult<object>(null);
             }
             else
             {
+                t = new Task(action);
                 t.Start(this);
             }
 
             return t;
         }
+
+        /// <summary>
+        /// Provides a completed task with the result for a syncronously run action.
+        /// this only needed for .NET40.  Depricated by <see cref="Task.FromResult()"/> in 4.5+
+        /// </summary>
+        /// <typeparam name="TResult"></typeparam>
+        /// <param name="result"></param>
+        /// <returns></returns>
+        private static Task<TResult> FromResult<TResult>(TResult result)
+        {
+            var tcs = new TaskCompletionSource<TResult>();
+            tcs.SetResult(result);
+            return tcs.Task;
+        }
+
 #else
         private void Run(Action action)
         {
@@ -140,7 +153,7 @@ namespace NetMQ
         }
 #endif
 
-        #endregion
+#endregion
 
         public NetMQPoller()
         {
--- a/src/NetMQ/NetMQPoller.cs
+++ b/src/NetMQ/NetMQPoller.cs
@@ -269,28 +269,11 @@ namespace NetMQ
 
         public void RemoveAndDispose<T>(T socket) where T : ISocketPollable, IDisposable
         {
-            if (socket == null)
-                throw new ArgumentNullException(nameof(socket));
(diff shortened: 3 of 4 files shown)


src/NetMQ.Tests/NetMQPollerTest.cs
```diff
@@ -400,6 +402,150 @@ public void RemoveSocket()
             }
         }
 
+        [Fact]
+        public async void RemoveAndDisposeSocket()
```
jasells: These two new tests could be refined a little, and at least a helper method to do the boiler plate setup in on place...

somdoron:
I think this is a breaking change, as you are changing a return type.
@drewnoakes what do you think?

jasells:
I didn't realize Run was public.  BUT, while technically breaking change from the standpoint of the public API changed, any code written to a void return type will be unaffected.

This is a much better option than leaving a thread race bug in, IMO.

It is also much better practice to return Task from async methods to avoid this very scenario in the first place, and the caller can decide whether they need to await completion of async task or not.

src/NetMQ/ISocketPollableCollection.cs
```diff
@@ -1,4 +1,5 @@
 using System;
+using System.Threading.Tasks;
```
jasells: Can be removed

src/NetMQ/NetMQPoller.cs
```diff
@@ -112,21 +113,47 @@ protected override void QueueTask(Task task)
             m_tasksQueue.Enqueue(task);
         }
 
-        public void Run([NotNull] Action action)
+        public Task Run([NotNull] Action action)
```
drewnoakes: Breaking public API is not a good idea, and goes against the [C4.1 process](https://rfc.zeromq.org/spec:22/C4/) that this project uses.

I agree that returning a task would be better. It'd be better to create overloads:

```c#
public Task RunAsync([NotNull] Action action);
public Task RunAsync([NotNull] Func<CancellationToken> action, CancellationToken cancellationToken);
```

src/NetMQ/NetMQPoller.cs
```diff
@@ -112,21 +113,47 @@ protected override void QueueTask(Task task)
...
+            {
                 action();
+
+                t = FromResult<object>(null);
+            }
             else
-                new Task(action).Start(this);
+            {
+                t = new Task(action);
+                t.Start(this);
```
drewnoakes: `Task.Factory.StartNew` is preferred over `new Task(...).Start` ([article](https://devblogs.microsoft.com/pfxteam/task-factory-startnew-vs-new-task-start/)).

src/NetMQ/NetMQPoller.cs
```diff
@@ -225,6 +252,11 @@ public void Remove(ISocketPollable socket)
         {
             if (socket == null)
                 throw new ArgumentNullException(nameof(socket));
+
+            // JASells: not sure I agree with this thow.
+            // If trying to remove a disposed socket, why complain?  It *might* work.  The issue
+            // is if the poller's thread tries to actually service the socket before the remove call...
+
```
drewnoakes: I'd remove this from the code and discuss in an issue/PR instead where conversation can happen more easily.

The rationale here was to help people discover when they're tearing things down in the wrong order, as that is a common cause of deadlocks and other problems. If you're going to fail some of the time, it's better to fail all of the time as it makes it easier to write code correctly.

src/NetMQ/NetMQPoller.cs
```diff
@@ -243,33 +275,18 @@ public void Remove(ISocketPollable socket)
                 socket.Socket.EventsChanged -= OnSocketEventsChanged;
                 m_sockets.Remove(socket.Socket);
                 m_isPollSetDirty = true;
-            });
+            })
+            .Wait();
```
drewnoakes: Calling `Wait` on tasks can lead to deadlocks in complex code.

I'd suggest making a `RemoveAsync` overload that returns `Task`.

As an aside, it's better to use `GetAwaiter().GetResult()` than `Wait()`, mostly because the latter wraps any thrown exception in a redundant `AggregateException` which can make debugging more difficult. Still, synchronous blocking waits in library code like this should be avoided. There are definitely user scenarios where deadlocks can occur.

src/NetMQ/NetMQPoller.cs
```diff
@@ -243,33 +275,18 @@ public void Remove(ISocketPollable socket)
...
+            //call the remove method
+            Remove(socket);
 
-                socket.Socket.EventsChanged -= OnSocketEventsChanged;
-                m_sockets.Remove(socket.Socket);
-                m_isPollSetDirty = true;
-                socket.Dispose();
-            });
+            //dispose of socket
+            socket.Dispose();
```
drewnoakes: This refactoring only works because of the synchronous wait, which isn't something we can safely do unfortunately.

If you add a `RemoveAndDisposeAsync` overload, its implementation could be simpler by awaiting `RemoveAsync` before disposing.

src/NetMQ/NetMQPoller.cs
```diff
@@ -280,7 +297,7 @@ public void Remove([NotNull] NetMQTimer timer)
 
             timer.When = -1;
 
-            Run(() => m_timers.Remove(timer));
+            Run(() => m_timers.Remove(timer)).Wait();
```
drewnoakes: Again, synchronous waits are not safe. This would have to be a `RemoveAsync` overload.

src/NetMQ/NetMQPoller.cs
```diff
@@ -293,7 +310,8 @@ public void Remove([NotNull] Socket socket)
             {
                 m_pollinSockets.Remove(socket);
                 m_isPollSetDirty = true;
-            });
+            })
+            .Wait();
```
drewnoakes: Ditto.

src/NetMQ.Tests/NetMQPollerTest.cs
```diff
@@ -400,6 +402,150 @@ public void RemoveSocket()
...
+            //allow for poller to continue running
+            await Task.Delay(2000);
+
+            patient.Stop();
+            Assert.False(patient.IsRunning);
+
+            canceller.Cancel();
+
+            pub?.Dispose();
+            patient?.Dispose();
```
drewnoakes: How could these be `null`?

src/NetMQ.Tests/NetMQPollerTest.cs
```diff
@@ -400,6 +402,150 @@ public void RemoveSocket()
...
+
+            //allow a little time to run
+            await Task.Delay(2000);
+
+            //now try to remove the sub from poller
+            patient.Remove(sub);
+
+            // dispose the sub (this will cause exception on poller's worker-thread) and it can't be caught!
+            sub.Dispose();
+            sub = null;
```
drewnoakes: Why set this to `null`?

src/NetMQ.Tests/NetMQPollerTest.cs
```diff
@@ -400,6 +402,150 @@ public void RemoveSocket()
...
+            //allow for poller to continue running
+            await Task.Delay(2000);
+
+            patient.Stop();
+            Assert.False(patient.IsRunning);
+
+            canceller.Cancel();
+
+            pub?.Dispose();
+            patient?.Dispose();
```
drewnoakes: How could these be `null`?

drewnoakes: `canceller` should be disposed as well.

src/NetMQ.Tests/NetMQPollerTest.cs
```diff
@@ -400,6 +402,150 @@ public void RemoveSocket()
...
+                while(!token.IsCancellationRequested)
+                {
+                    pub.SendFrame($"Hello-{++i}");
+
+                    // send ~ 5Hz
+                    await Task.Delay(200);
+                }
+            });
+
+            var pubThread = Task.Run(pubAction);
```
drewnoakes:
```suggestion
            Task.Run(pubAction);
```

src/NetMQ.Tests/NetMQPollerTest.cs
jasells: 1) it doesn't hurt to null check
2) I often use the convention of setting disposed objects to null after calling dispose().  Just a personal habit that allows easy check obj?. instead of object.IsDisposed().
3) allows for initialization/re-initialization from same code block (if set = null after dispose convention is used)

src/NetMQ.Tests/NetMQPollerTest.cs
jasells: see above [comment](https://github.com/zeromq/netmq/pull/835/files/a3819e1c51437784b7c7f2aab221e52058cf0bf5#r360645822)

src/NetMQ/NetMQPoller.cs
jasells: ahh.... yes, I was wishing ```Thread.Start()``` was fluent.

src/NetMQ/NetMQPoller.cs
jasells: Yes, this is a note to do that =)

src/NetMQ/NetMQPoller.cs
jasells: 100% agree, but was trying to avoid that because I originally thought only the ```Remove`` methods were public, and chaining only the non-public ```RunAsync``` method would avoid breaking changes...

Why is ```RunAsync``` even public?

The only reason I can see is to work around the original thread race bug when calling ```Remove()```?

src/NetMQ/NetMQPoller.cs
jasells: Lol.

@ first i tried ```await```, but there was an issue using ```await``` in .NET40....  so I changed it to Wait()....  forgot about these methods.

drewnoakes: The problem likely was that you cannot `await` in a non-`async` method.

src/NetMQ/NetMQPoller.cs
drewnoakes: > Why is `RunAsync` even public?

`NetMQPoller` is a `TaskScheduler` and can be used to schedule work to run on the same synchronisation context (in this case: thread) as the poller. `RunAsync` is useful when you need to schedule work that needs to be on that thread, which is not an uncommon requirement.

Regardless, it's public API and must be preserved.

src/NetMQ/NetMQPoller.cs
drewnoakes: Even if `Thread.Start` did exist, it would be too heavyweight. We don't need or want to start a new thread for this background work. We need this task to run asynchronously on the poller's thread, which is achieved by passing `this` as the `Task`'s `TaskScheduler`.

jasells: Sry, I said `Thread.Start `but I meant `Task.Start()`

jasells:
I haven't forgotten about this PR. Been busy with holidays and getting back to work.  I will try to address the concerns raised here this week and issue an update.

src/NetMQ/NetMQPoller.cs
jasells: > Regardless, it's public API and must be preserved.

Got that.

> RunAsync is useful when you need to schedule work that needs to be on that thread, which is not an uncommon requirement.

I think a public `RunAsync` is a design flaw.  NetMQ objects, `NetMQPoller ` in this case could and *should* handle marshalling work that *needs* to be done on the polling thread internally.  This exposes a large path for external code to break  `NetMQPoller ` and is the root source of the issue being discussed.  The only case I can think of currently *is this* case (add/remove a socket from the polled group).  Certainly, the *primary* use case for this method is the async polling of the sockets in the list for data... for which all the rest of the framework code is either internal to NetMQ or protected/private, which makes sense.

src/NetMQ/NetMQPoller.cs
```diff
@@ -112,21 +113,47 @@ protected override void QueueTask(Task task)
...
+        /// </summary>
+        /// <typeparam name="TResult"></typeparam>
+        /// <param name="result"></param>
+        /// <returns></returns>
+        private static Task<TResult> FromResult<TResult>(TResult result)
+        {
+            var tcs = new TaskCompletionSource<TResult>();
+            tcs.SetResult(result);
+            return tcs.Task;
+        }
```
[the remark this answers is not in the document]
jasells: I cannot find a way to do this currently, as there are already several conditional blocks that will require significant re-organizing (hopefully that's all that is required) to wrap this in the correct conditional.

src/NetMQ/NetMQPoller.cs
```diff
@@ -112,21 +113,47 @@ protected override void QueueTask(Task task)
...
+
 #else
         private void Run(Action action)
         {
             action();
         }
 #endif
 
-        #endregion
+#endregion
```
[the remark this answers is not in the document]
jasells: Should be addresses in latest.

src/NetMQ.Tests/NetMQPollerTest.cs
```diff
@@ -400,6 +402,150 @@ public void RemoveSocket()
...
+            Assert.False(patient.IsRunning);
+
+            canceller.Cancel();
+
+            pub?.Dispose();
+            patient?.Dispose();
+        }
+
+        [Fact]
+        public async void DisposeSocketAfterRemoval()
```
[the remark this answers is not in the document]
jasells: I think I addressed the concerns in this thread...

src/NetMQ/NetMQPoller.cs
```diff
@@ -112,21 +113,47 @@ protected override void QueueTask(Task task)
...
-        public void Run([NotNull] Action action)
+        public Task Run([NotNull] Action action)
         {
+            Task t;
+
             if (!IsRunning || CanExecuteTaskInline)
+            {
                 action();
+
+                t = FromResult<object>(null);
```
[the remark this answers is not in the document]
jasells: This same pattern is also in existing overloaded method `ContainsAsync`.  Also, I'm not convinced it is a good idea to share `TaskCompletionSource` instances among (potentially) multiple threads...?

src/NetMQ/NetMQPoller.cs
jasells: I think the latest addresses this by adding new `RemoveAsync()` method, and reverting the old version, along with the original issue.
I marked the `Remove` methods as `Obsolete`...

src/NetMQ/NetMQPoller.cs
jasells: In latest.

jasells:
OK, so I pushed up several new commits to (hopefully) address the breaking API concerns, while still fixing the sync issues by *adding* `Task RemoveAsyc()` and `Task RunAsync()` methods and leaving the existing API intact (though still broken, if used).

src/NetMQ/NetMQPoller.cs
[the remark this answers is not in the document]
jasells: See this [commit ](https://github.com/zeromq/netmq/pull/835/commits/455f6b066c6a775f91e51e4a2c88879bbe2b86c5) which I think addresses this concern.

src/NetMQ/NetMQPoller.cs
jasells: should be resolved in latest.

jasells:
Hopefully, resolved all conflicts, and concerns...

jasells:
So, just checking in.  Haven't seen any response in 10 days.

src/NetMQ/NetMQPoller.cs
```diff
@@ -120,25 +120,63 @@ protected override void QueueTask(Task task)
...
+            }
+
+            return t;
+        }
+
         /// <summary>
         /// Run an action on the Poller thread
         /// </summary>
         /// <param name="action">The action to run</param>
+        [Obsolete("Queues the action on the poller's thread, but provides no sync mechanism.  Please use RemoveAsync() to remove sockets or timers")]
```
drewnoakes: This message needs reviewing.

src/NetMQ/NetMQPoller.cs
```diff
@@ -120,25 +120,63 @@ protected override void QueueTask(Task task)
...
+
+            if (!IsRunning || CanExecuteTaskInline)
+            {
+                action();
+
+                t = FromResult<object>(null);
+            }
+            else
+            {
+                t = Task.Factory.StartNew(action);
```
drewnoakes: The `Run` method's purpose was to execute `action` on the poller's thread. It did this by providing `this` as the task scheduler.

This new `RunAsync` method does not guarantee the action runs on the poller's thread.

src/NetMQ/NetMQPoller.cs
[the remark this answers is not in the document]
drewnoakes: A completed task is safe to share. Use `Task.FromResult<object>(null)` for TFMs that support it, and cache your own (from a `TaskCompletionSource<>`) in remaining TFMs.

src/NetMQ/NetMQPoller.cs
```diff
@@ -259,15 +297,27 @@ public void Add([NotNull] Socket socket, [NotNull] Action<Socket> callback)
...
+        }
+
+        public Task RemoveAsync(ISocketPollable socket)
         {
             if (socket == null)
                 throw new ArgumentNullException(nameof(socket));
+
+            // JASells: not sure I agree with this thow.
+            // If trying to remove a disposed socket, why complain?  It *might* get removed before the poller thread accesses it.  
+            // The issue is if the poller's thread tries to actually service the disposed socket before the remove call...
```
drewnoakes: I believe the intent here was to reliably fail when things aren't torn down in the correct order. IIRC, by not throwing here it was possible that code would only fail _some_ of the time, which (perhaps counterintuitively) is worse than failing all of the time.

src/NetMQ/NetMQPoller.cs
```diff
@@ -120,25 +120,63 @@ protected override void QueueTask(Task task)
...
+
+#else  //NET35
+            private void Run(Action action)
         {
             action();
         }
 #endif
 
-        #endregion
+    #endregion
```
drewnoakes: Indentation here needs reviewing.

src/NetMQ/NetMQPoller.cs
```diff
@@ -120,25 +120,63 @@ protected override void QueueTask(Task task)
...
         /// Run an action on the Poller thread
         /// </summary>
         /// <param name="action">The action to run</param>
+        [Obsolete("Queues the action on the poller's thread, but provides no sync mechanism.  Please use RemoveAsync() to remove sockets or timers")]
         public void Run([NotNull] Action action)
         {
             if (!IsRunning || CanExecuteTaskInline)
                 action();
             else
                 new Task(action).Start(this);
```
drewnoakes: I think this should use `Task.Factory.StartNew`.

src/NetMQ/NetMQPoller.cs
drewnoakes: I'm not sure we should force people to use the new async APIs. What do you think @somdoron?

src/NetMQ/NetMQPoller.cs
```diff
@@ -225,7 +263,7 @@ public void Add([NotNull] NetMQTimer timer)
                 throw new ArgumentNullException(nameof(timer));
             CheckDisposed();
 
-            Run(() => m_timers.Add(timer));
+            RunAsync(() => m_timers.Add(timer));
```
drewnoakes: For completeness, shouldn't we have `AddAsync` methods that return tasks? These sync methods could call into them and discard the `Task`.

Ditto other sync methods such as `Remove`.

src/NetMQ/NetMQPoller.cs
```diff
@@ -330,7 +387,7 @@ public void Remove([NotNull] NetMQTimer timer)
 
             timer.When = -1;
 
-            Run(() => m_timers.Remove(timer));
+            RunAsync(() => m_timers.Remove(timer));
```
drewnoakes: Is there a reason you didn't create `RemoveAsync(NetMQTimer)`?

src/NetMQ/NetMQPoller.cs
```diff
@@ -344,10 +401,13 @@ public void Remove([NotNull] Socket socket)
                 throw new ArgumentNullException(nameof(socket));
             CheckDisposed();
 
-            Run(() =>
+            RunAsync(() =>
```
drewnoakes: Is there a reason you didn't create `RemoveAsync(Socket)`?

src/NetMQ/NetMQPoller.cs
```diff
@@ -344,10 +401,13 @@ public void Remove([NotNull] Socket socket)
...
             {
-                m_pollinSockets.Remove(socket);
-                m_isPollSetDirty = true;
+                if (m_pollinSockets.Keys.Contains(socket))
+                {
+                    m_pollinSockets.Remove(socket);
+                    m_isPollSetDirty = true;
+                }
```
drewnoakes: `Dictionary<,>.Remove` returns `bool`, so you can avoid the double-lookup by removing the `Contains` call.

Also, FYI `.Keys.Contains` calls `Enumerable.Contains` which has O(N) performance. If you want to test whether a dictionary contains a key, use `ContainsKey` instead which has O(1) performance.

src/NetMQ/NetMQPoller.cs
```diff
@@ -369,7 +429,7 @@ public Task<bool> ContainsAsync([NotNull] ISocketPollable socket)
             CheckDisposed();
 
             var tcs = new TaskCompletionSource<bool>();
-            Run(() => tcs.SetResult(m_sockets.Contains(socket)));
+            RunAsync(() => tcs.SetResult(m_sockets.Contains(socket)));
             return tcs.Task;
```
drewnoakes: Do you need the TCS here? Create `Task<T> RunAsync<T>(Func<T>)` and delegate to it instead.

Ditto the other `ContainsAsync` overloads.

jasells:
For some reason, I didn't get any notifications about the comments.  I'll look into them over the next few days.

src/NetMQ/NetMQPoller.cs
jasells: I can make that change, but how many sockets are really going to be in this collection? I doubt enough to matter, nor is this likely to be called often.

src/NetMQ/NetMQPoller.cs
jasells: oops.  You're right.  Will fix.

src/NetMQ/NetMQPoller.cs
jasells: This doesn't force anyone to use the new API's, it just forces them to opt-out (with compiler ignores for release build).

It is strongly suggesting to use the new API's, but that is much better than the status-quo, IMHO: can't dispose sockets at all, just have to rely on GC.

src/NetMQ/NetMQPoller.cs
jasells: I did, line [307.](https://github.com/zeromq/netmq/blob/455f6b066c6a775f91e51e4a2c88879bbe2b86c5/src/NetMQ/NetMQPoller.cs#L307).

The crrent implementation was intentionally done to keep current behavior when calling the public API's original methods.  The only change from master I made here was `Run()`->`RunAsync()` in an effort to minimize my changes. 

We *could* call `RemoveAsync(socket)` here, to remove a secondary call chain? That's your call as the owner.  I think it is probably a good idea...

src/NetMQ/NetMQPoller.cs
jasells: I didn't run into any problems with the current add implementations...

BUT since you ask, it is *probably* possible to do 
```
poller.Add(socket);
await poller.RemoveAsync(socket);
socket.Dispose();
```

and still have issues?  Maybe?  It is possibly a race condition?

src/NetMQ/NetMQPoller.cs
jasells: I was just focused on the sockets, but that should follow the same pattern.  Once we get the socket-related updates to everyone's liking, we should be able to copy for  `NetMQTimer`?

src/NetMQ/NetMQPoller.cs
jasells: Ahh... yes, I see.  I was thinking of the indexer `dictionary[key]` .  It is unnecessary.

jasells: I think I added it trying to track down the error I saw that I ultimately tracked to `NetMQSelector`.

jasells: removed the check

src/NetMQ/NetMQPoller.cs
jasells: Again, my only change here is `Run(action)` to `RunAsync(action)`.

We *could* add the suggested `Task<T> RunAsync<T>(Func<T>)` overload... but I would striongly suggest it should be `private` or at least `internal`.

src/NetMQ/NetMQPoller.cs
jasells: This is identical to the code in the original `Run(Action)` method.  See [here](https://github.com/zeromq/netmq/blob/455f6b066c6a775f91e51e4a2c88879bbe2b86c5/src/NetMQ/NetMQPoller.cs#L146).

Using `Task.Factory.StartNew` requires several additional parameters in addition to specifying the scheduler: 

[`Task<TResult> StartNew<TResult>(Func<TResult> function, CancellationToken cancellationToken, TaskCreationOptions creationOptions, TaskScheduler scheduler)`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskfactory.startnew?view=netframework-4.8#System_Threading_Tasks_TaskFactory_StartNew_System_Action_System_Threading_CancellationToken_System_Threading_Tasks_TaskCreationOptions_System_Threading_Tasks_TaskScheduler_)

Certainly possible, but I'm uncertain what the `TaskCreationOptions` should be here, nor do we supply a `CancellationToken`.

Adding a `cancellationToken` parameter would break the existing API, and defining one requires creating it even though  it won't be used to stop the task, unless an arbitrary timeout is assigned.

src/NetMQ/NetMQPoller.cs
jasells: I belive the fix I made [here ](https://github.com/jasells/netmq/blob/9f17258c917a94422a87a7532aea3f29b3d0b58c/src/NetMQ/NetMQSelector.cs#L169)is 1/2 of the required fix to be able to alleviate this concern entirely.

Would need to do the same check [here](https://github.com/jasells/netmq/blob/9f17258c917a94422a87a7532aea3f29b3d0b58c/src/NetMQ/NetMQSelector.cs#L130).

jasells: Sorry forgot I agreed to separate this...  I will do that ASAP, so as to keep things cleaner.

src/NetMQ/NetMQPoller.cs
jasells: Deleted first response, I misread.  You are talking about `System.Net.Sockets.Socket` type here.  I didn't even realize it was available...  So that's why I didn't add the method.

Never occurred to me to use a raw socket with this lib...

src/NetMQ/NetMQPoller.cs
jasells: Removed this comment from source.

src/NetMQ/NetMQPoller.cs
jasells: Added overloaded `RemoveAsync()`

src/NetMQ/NetMQPoller.cs
jasells: added overload and tesst

src/NetMQ/NetMQPoller.cs
jasells: added suggested `Task<T> RunAsync<T>(Func<T>)`

jasells:
OK, so does this PR have a chance, or are you guys ignoring it.  If the latter, I'll just fork and move on with my life.

jzi96:
Any update to this branch and issue? Issue is closed but branch not merged?

Thx

drewnoakes:
A bunch of the feedback I left was not addressed. Even still, there's a non-trivial amount of work in reviewing this change for correctness and pandemic life isn't giving me much free time.

jzi96:
Thanks for the feedback Drew. Good to know. :)

Stay save,
Jan

jasells:
Is the only remaining issue the 'AddAsync' comment?  I replied, but never got a response.  In case it is unclear, I see no need for AddAsync, but they could be added, I suppose.   I'll have to refresh my memory at this point, and it looks like there are conflicts from master updates.

src/NetMQ/NetMQPoller.cs
jasells: Is this a sticking point?  It can easily be removed, but is not my preference...  But need clear direction from NetMq team as to the final yay or nay.

What is the preferred method to communicate this issue and change to application coders?

drewnoakes:
@jasells for me at least, all unresolved comments are still issues for me. In each case my original comment stands.

src/NetMQ/NetMQPoller.cs
jasells: @drewnoakes I need further clarification here.   This request seems to require creating an API change?  The current `Task(action).Start(this)` ensures that the task is run on the poller thread.

Is there a compatibility issue with the current impl?

src/NetMQ/NetMQPoller.cs
jasells: @drewnoakes So, I should interpret "I'm not sure..."  as "I'm not approving this until it removed" ?

src/NetMQ/NetMQPoller.cs
jasells: Ahh, I see.  There is possibly some confusion in using `Task(action).Start(this)` vs. using the locally defined `Run(action)` method [here](https://github.com/zeromq/netmq/blob/e86706ffb9138a25b3bc0e10484efe8b89a64819/src/NetMQ/NetMQPoller.cs#L482).

These are functionally equivalent, but I can update this code to use the local `Run(action)` method for consistency?

PR #839, opened 2019-12-20
xpub/xsub documentation update

eatfrog:
instead of using @/> as connectionstring prefixes, make it clear that one needs to use bind and the others connect


somdoron:
Thanks

PR #843, opened 2020-01-13
Zmtp3

somdoron:


somdoron:
@drewnoakes this is ready to be reviewed and merged

src/NetMQ/Core/Utils/ByteArrayUtility.cs
```diff
@@ -0,0 +1,22 @@
+namespace NetMQ.Core.Utils
+{
+    public static class ByteArrayUtility
```
drewnoakes: Potentially make this `internal`.

somdoron: fixed

somdoron:
Thanks, this is actually a preparation for the curve encryption.

PR #844, opened 2020-01-17
Publish is leaking msgs because of shared msg bug

### LINKED ISSUE #739: dotnet core InitPool, XPub/Publish, and reference counting leading to bad data race condition
### Environment

    NetMQ Version:   4.0.0.1
    Operating System: Windows 7
    .NET Version:  .net core

### Expected behaviour
Two topics are being sent,  "Logging_test" and "TestTopic"
topicMessage.InitPool(topicLength);
topicMessage.Put(topic, 0, topicLength);
 _publisher.Send(ref topicMessage, true);
topicMessage.Close();

payloadMessage.InitPool(payloadLength);
_publisher.Send(ref payloadMessage, false);
payloadMessage.Close();

Expect that only those topics and a payload would be sent. This happened the vast majority of the time, until it didn't, usually under load. 

### Actual behavior

On the receiving end, I would get the proper values 99.99% of the time, but sometimes I would get a topic that didn't match its size. 
IE: expected "Logging_test\n" but would receive "TestTopic\nest\n sometimes. with a size of 15 instead of 10.
Reason seemed to be the reference counting on the msg structure. Basically the "data" would go back into the pool before the msg was actually freed up.   

note, using InitGC this did not happen at all, which further lead me to believe it was the reference counting

send/receive from a pub/sub with a small topic/payload in rapid succession, create dictionaries to look at unique keys received.  Create load by doing other things within the send process. 

### Possible steps to correct the issue (or at least what worked for me)

I can setup a pull request but want to make sure I am barking up the right tree.
I have made two small modifications that have seemed to correct the issue. 
In, src/NetMQ/Msg.cs
``` csharp
            public void AddReferences(int amount)
        {
            if (amount == 0)
            {
                return;
            }

            if (MsgType == MsgType.Pool)
            {
                if (IsShared)
                {
                    m_refCount.Increase(amount);
                }
                else
                {
                    m_refCount.Increase(amount); //<=change from set to Increase. as that seems to be with the main function is for.
                    Flags |= MsgFlags.Shared;
                }
            }
        }
```
in src/NetMQ/Core/Patterns/Utils/Distribution.cs
``` csharp
        private void Distribute(ref Msg msg)
        {
            // If there are no matching pipes available, simply drop the message.
            if (m_matching == 0)
            {
                msg.Close();
                msg.InitEmpty();

                return;
            }

            if (msg.MsgType != MsgType.Pool)
            {
                for (int i = 0; i < m_matching; ++i)
                {
                    if (!Write(m_pipes[i], ref msg))
                    {
                        --i; //  Retry last write because index will have been swapped
                    }
                }

                msg.Close();
                msg.InitEmpty();

                return;
            }

	//when a msg is created, it actually has zero references, as it doesn't count its own reference. 
	//add to the current based off the number of copies we had to create.  add to the reference so that
	//when the class counter hits zero we can return the buffers.
            msg.AddReferences(m_matching); //<== modified here to remove the -1, as it didn't seem to line up with the msg structure atomiccounter as it starts at zero, with only copies having a reference count.

            // Push copy of the message to each matching pipe.
            int failed = 0;
            for (int i = 0; i < m_matching; ++i)
            {
                if (!Write(m_pipes[i], ref msg))
                {
                    ++failed;
                    --i; //  Retry last write because index will have been swapped
                }
            }
            if (failed != 0)
                msg.RemoveReferences(failed);

            // Detach the original message from the data buffer. Note that we don't
            // close the message, as the original message doesn't count as a reference. we want to init here, in case they "close" so it doesn't manipulate the counter.
            msg.InitEmpty(); 
        }
```

somdoron:
fixes #739

### DIFF
--- a/src/NetMQ/Msg.cs
+++ b/src/NetMQ/Msg.cs
@@ -276,7 +276,10 @@ namespace NetMQ
                 }
                 else
                 {
-                    m_refCount.Set(amount);
+                    // Because msg is not yet shared we add 1 to the amount to represent the current copy of the msg.
+                    // We can set the amount in a none thread-safe way because the message is not yet shared and
+                    // therefore only being used by one thread
+                    m_refCount.Set(amount + 1);
                     Flags |= MsgFlags.Shared;
                 }
             }


somdoron:
Actually it doesn't need to be thread-safe yet, when the msg is not shared it means it not yet thread-safe. Only after the msg is shared it becomes thread-safe.
It is actually a bug of porting the class from libzmq, which got it correctly:
https://github.com/zeromq/libzmq/blob/master/src/msg.cpp#L501

Anyway, I will add a comment,

somdoron:
@drewnoakes added a comment

PR #848, opened 2020-01-25
Apply codecov

### LINKED ISSUE #784: Suggest to append codecov
Suggest to append codecov for measure code coverage

moreal:
In this PR, *codecov* was applied to resolve #784, with reference [codecov/example-csharp].

I expect the stats of coverage will be shown in README like below. 

image

It is only tested in my environment, so it is not clear if it will work. but it expects to work well.

[codecov/example-csharp]: https://github.com/codecov/example-csharp


somdoron:
Thanks!

somdoron:
Now we need to increase the coverage...

moreal:
> Now we need to increase the coverage...

Yes.. it needs to increase the coverage. In my thought, the codecov should be added into required checks (e.g., *appveyor build*, *CodeFactor*, ...) to stop the decreasement. How do you think about it? @somdoron

PR #851, opened 2020-02-04
Problem: NetMQ is not supporting curve encryption

somdoron:


mtspencer:
somdoron - I've just pulled this latest commit in order to try out the curve encryption in our system.  From what I can tell running some minimal tests it seems to be working very well for us.  Would you be able to push a pre-release version to Nuget in the near future?

somdoron:
Amazing, thank you for testing it out.
I will push a version early next week.

mtspencer:
Excellent, I'll continue testing with my local build and let you know if I see any issues

somdoron:
https://www.nuget.org/packages/NetMQ/4.0.0.352-pre

src/NetMQ.Tests/CurveTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            using var server = new DealerSocket();
+            server.Options.CurveServer = true;
+            server.Options.CurveCertificate = serverPair;
+            server.Bind($"tcp://127.0.0.1:55367");
+            
+            var clientPair = new NetMQCertificate();
+            using var client = new DealerSocket();
+            client.Options.CurveServerKey = serverPair.PublicKey;
+            client.Options.CurveCertificate = clientPair;
+            client.Connect("tcp://127.0.0.1:55367");
```
davide-costa: @somdoron  I would like to use curve encryption in a scenario where i need to ensure integrity, authenticity and confidentiality. For what I understood in this example/test only server identity is verified, is it possible to verify also clients'? In fact, in my scenario there isn't a client-server architecture but clients/peers communicating with each other.

Thanks in advance

somdoron: yes, however that is not implemented yet.
We need to port ZAP from libzmq for that. I will take some more time.

somdoron: You can develop your own protocol on top. Either with username and password or EC signing.

davide-costa: Thanks

Issue #849, opened 2020-01-26
Increase code coverage


moreal:
Now we need to increase the coverage...

_Originally posted by @somdoron in https://github.com/zeromq/netmq/pull/848#issuecomment-578403651_

I agree for it. In my thought, the *codecov* should be added into required checks (e.g., appveyor build, CodeFactor, ...) to stop the decreasement of the coverage.

somdoron:
Sounds good to me

moreal:
Then could you add it after observe other opinions for times? Because I am not maintainer and have no permissions for this repository.
IMHO: It seems not sure to collect enough opinions to say that the opinions was came and gone, from other users. Or can it be decided only with maintainers' confirms?

moreal:
Yeah, I expect it will help stay code stable 😄. Is this needed more discussion? or it's ready to apply?

moreal:
It seems already applied. Is it better to close this? Or is it better to leave it open for reminding? Tell me your opinions.

somdoron:
Lets leave it open. If you want to change anything at the project level let
me know and I would give you a permission

moreal:
I'm sorry to delete the comment because it was not really applied yet. Now there are some checks like below.

![image](https://user-images.githubusercontent.com/26626194/73958340-24620580-494b-11ea-9d51-f3a9d38bb685.png)

But I also want to add *codecov* as checks like below. because if not, it can be merged without check ing code coverage of it.

![image](https://user-images.githubusercontent.com/26626194/73960390-27122a00-494e-11ea-899d-391e43e99b23.png)

Further more, I want to suggest to set up an issue management strategy and pre-release. Also I'm sorry but I couldn't understand completely what the *permission* means there.. Could you let me know more about it please?

moreal:
If there was some parts I didn't understood well, let me know that 😁 @somdoron

somdoron:
@moreal I'm not sure what I need to do in order to add the codecov, can you link to a tutorial or something?

moreal:
There is helpful document provided from codecov, [here][codecov-docs-link]. I'm not sure but it might require to install [codecov github app] on this repository! If there was hard to set up and the document is not enough, I'll try to write simple tutorial with gist or screen capture. Or it's just thought, if you can give me a temporary permission, I could try it.

[codecov github app]: https://github.com/apps/codecov/
[codecov-docs-link]: https://docs.codecov.io/docs/commit-status

somdoron:
[...]

PR #854, opened 2020-02-12
Configure codecov

moreal:
In #849, it was suggested to use codecov as commit status. Therefore in this pull request, *codecov.yml*, codecov configuration file was added. After this pull request was merged, it expects to work well and to see it between commit statuses. All pull requests will become to increase code coverage.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,12 @@
+coverage:
+  range: 65..95
+  status:
+    project:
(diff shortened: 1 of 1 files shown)


moreal:
`ChangeTimerInterval` test was failed in *appveyor CI*, seems non deterministic. It should be fixed.

Issue #872, opened 2020-05-12
PgmSender Encoding Exception


mikepmiller:
### Environment

    NetMQ Version: Latest Build in Master (specifically the commit https://github.com/zeromq/netmq/commit/6726e07b1dc4dd77283a3e0ec5edf0e03b454f32)
    Operating System: Windows 10
    .NET Version: 4.7.1

### Expected behaviour
Sending messages via PublisherSocket over PGM shouldn't crash due to an ArgumentException.  Specifically happens when there is more than a fully buffer (1500 bytes) of data waiting to be published.

### Actual behaviour
An exception is thrown:
  

> System.ArgumentException
>   HResult=0x80070057
>   Message=Offset and length were out of bounds for the array or count is greater than the number of elements from index to the end of the source collection.
>   Source=mscorlib
>   StackTrace:
>    at System.Buffer.BlockCopy(Array src, Int32 srcOffset, Array dst, Int32 dstOffset, Int32 count)
>    at NetMQ.Core.Transports.ByteArraySegment.CopyTo(Int32 fromOffset, ByteArraySegment dest, Int32 destOffset, Int32 toCopy) in C:\source\netmq\src\NetMQ\Core\Transports\ByteArraySegment.cs:line 341
>    at NetMQ.Core.Transports.EncoderBase.Encode(ByteArraySegment& data, Int32 size) in C:\source\netmq\src\NetMQ\Core\Transports\EncoderBase.cs:line 131
>    at NetMQ.Core.Transports.Pgm.PgmSender.BeginSending() in C:\source\netmq\src\NetMQ\Core\Transports\Pgm\PgmSender.cs:line 227
>    at NetMQ.Core.Transports.Pgm.PgmSender.ActivateOut() in C:\source\netmq\src\NetMQ\Core\Transports\Pgm\PgmSender.cs:line 132
>    at NetMQ.Core.SessionBase.ReadActivated(Pipe pipe) in C:\source\netmq\src\NetMQ\Core\SessionBase.cs:line 332
>    at NetMQ.Core.Pipe.ProcessActivateRead() in C:\source\netmq\src\NetMQ\Core\Pipe.cs:line 382
>    at NetMQ.Core.ZObject.ProcessCommand(Command cmd) in C:\source\netmq\src\NetMQ\Core\ZObject.cs:line 261
>    at NetMQ.Core.IOThread.Ready() in C:\source\netmq\src\NetMQ\Core\IOThread.cs:line 96
[...]

PR #873, opened 2020-05-12
PgmSender Encoding Fix

mikepmiller:
Previously buffer was shifted by _aggregate_ amount on every loop iteration; should only be shifted by number of bytes encoded on that pass.


mikepmiller:
Fixes #872

PR #874, opened 2020-05-12
PgmSession: Add handling for ConnectionAborted

mikepmiller:
I've seen this far less frequently than the ConnectionReset, but from what I can tell from the documentation, this is also not a fatal error.  Given that it's hard to catch exceptions thrown in the IOThread, I prefer calling Error and continuing execution.  For context, I've been using this change for several years and have never seen any negative consequences.  (Totally acknowledged that I may've suppressed a critical issue and just missed it!)

Thanks very much!


mikepmiller:
Thanks for merging so quickly!

PR #875, opened 2020-05-15
Enable SourceLink, deterministic builds, embedded icon and project file clean up

drewnoakes:


somdoron:
@drewnoakes I'm actually using github actions for releases now, so we will have to update the following files as well:
https://github.com/zeromq/netmq/blob/master/.github/workflows/release.yml
https://github.com/zeromq/netmq/blob/master/.github/workflows/prerelease.yml

PR #876, opened 2020-05-20
problem: no way to cancel receive operation

somdoron:
Solution: add cancellation token to all receive operation (thread safe only, for now)

### DIFF
--- a/src/NetMQ/Core/SocketBase.cs
+++ b/src/NetMQ/Core/SocketBase.cs
@@ -25,6 +25,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Net.Sockets;
+using System.Threading;
 using AsyncIO;
 using JetBrains.Annotations;
 using NetMQ.Core.Patterns;
@@ -982,6 +983,7 @@ namespace NetMQ.Core
         /// </summary>
         /// <param name="msg">the <c>Msg</c> to read the received message into</param>
         /// <param name="timeout">this controls whether the call blocks, and for how long.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None"/>.</param>
         /// <returns><c>true</c> if successful, <c>false</c> if it timed out</returns>
         /// <remarks>
         /// For <paramref name="timeout"/>, there are three categories of value:
@@ -993,7 +995,7 @@ namespace NetMQ.Core
         /// </remarks>
         /// <exception cref="FaultException">the Msg must already have been uninitialised</exception>
         /// <exception cref="TerminatingException">The socket must not already be stopped.</exception>
-        public bool TryRecv(ref Msg msg, TimeSpan timeout)
+        public bool TryRecv(ref Msg msg, TimeSpan timeout, CancellationToken cancellationToken = default)
         {
             Lock();
             try
@@ -1056,7 +1058,10 @@ namespace NetMQ.Core
                 bool block = m_ticks != 0;
                 while (true)
                 {
-                    ProcessCommands(block ? timeoutMillis : 0, false);
+                    if (cancellationToken.IsCancellationRequested)
+                        return false;
+                    
+                    ProcessCommands(block ? timeoutMillis : 0, false, cancellationToken);
 
                     isMessageAvailable = XRecv(ref msg);
                     if (isMessageAvailable)
@@ -1168,15 +1173,21 @@ namespace NetMQ.Core
         /// </summary>
         /// <param name="timeout">how much time to allow to wait for a command, before returning (in milliseconds)</param>
         /// <param name="throttle">if true - throttle the rate of command-execution by doing only one per call</param>
+        /// <param name="cancellationToken">allows the caller to cancel the process commands operation</param>
         /// <exception cref="TerminatingException">The Ctx context must not already be terminating.</exception>
-        private void ProcessCommands(int timeout, bool throttle)
+        private void ProcessCommands(int timeout, bool throttle, CancellationToken cancellationToken = default)
         {
             bool found;
             Command command;
             if (timeout != 0)
             {
-                // If we are asked to wait, simply ask mailbox to wait.
-                found = m_mailbox.TryRecv(timeout, out command);
+                if (cancellationToken.CanBeCanceled)
+                {
+                    using var registration = cancellationToken.Register(SendCancellationRequested);
+                    found = m_mailbox.TryRecv(timeout, out command);
+                }
+                else
+                    found = m_mailbox.TryRecv(timeout, out command);
             }
             else
             {
--- a/src/NetMQ/ReceiveThreadSafeSocketExtensions.cs
+++ b/src/NetMQ/ReceiveThreadSafeSocketExtensions.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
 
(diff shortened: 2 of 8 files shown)


src/NetMQ/ReceiveThreadSafeSocketExtensions.cs
```diff
@@ -88,15 +101,20 @@ public static bool TryReceiveBytes([NotNull] this IThreadSafeInSocket socket, ou
...
+            CancellationToken cancellationToken = default)
         {
             if (TryReceiveBytes(socket, out var bytes))
                 return new ValueTask<byte[]>(bytes);
 
             // TODO: this is a hack, eventually we need kind of IO ThreadPool for thread-safe socket to wait on asynchronously
             // and probably implement IValueTaskSource
-            return new ValueTask<byte[]>(Task.Factory.StartNew(socket.ReceiveBytes, TaskCreationOptions.LongRunning));
+            // TODO: should we avoid lambda here as it cause heap allocation for the environment?
+            return new ValueTask<byte[]>(Task.Factory.StartNew(() => socket.ReceiveBytes(cancellationToken),
```
somdoron, line 116: @drewnoakes do you think we would benefit here from de-lambda the action? using a real method with the cancellation token passed as a state object?

src/NetMQ/Core/SocketBase.cs
```diff
@@ -993,7 +995,7 @@ public bool TrySend(ref Msg msg, TimeSpan timeout, bool more)
         /// </remarks>
         /// <exception cref="FaultException">the Msg must already have been uninitialised</exception>
         /// <exception cref="TerminatingException">The socket must not already be stopped.</exception>
```
drewnoakes, line 997: Would be good to document `OperationCanceledException` here too, and a few other places in this file.

src/NetMQ/ReceiveThreadSafeSocketExtensions.cs
```diff
@@ -18,14 +19,24 @@ public static class ReceiveThreadSafeSocketExtensions
...
             msg.InitEmpty();
 
-            socket.Receive(ref msg);
+            try
+            {
+                socket.Receive(ref msg, cancellationToken);
+            }
+            finally
+            {
+                msg.Close();
```
drewnoakes, line 38: Is it safe to close it then use it below?

src/NetMQ/ReceiveThreadSafeSocketExtensions.cs
drewnoakes: It would avoid a heap allocation for the closure object, yes. You'd need to pass both the socket and token as state, then cast them at the other side.

I played around with this a little:

- [lambda](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABADAAgwRgDoAVAC1gEMATASwDsBzAbgFgAobPfAVlY4GY8AJhwBhdgG92OGXkEA1CgBsArjAxIAPASwA+HAFkAFMAgQlOYABoxFOmBhKlFAC40IdYhADWMOjhcfPwBKaVkpNlkonBoAMxwTUMjolIwAdhw6GAB3HEVVdS0dXSMAIhcoNVLgvhSZMNSMrNz8tQ1tfD0jDUIAMQowQKgAT0IAZRcKKBcAORyjI2CcAF59QN86QgBJAGdROwcnV3c6ACUYAEc1HZcYKhwAfhxSsAPHJTvSnBBn2OUdmDVGoNHAAX3YoKAA==)
- [method with state passing](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABADAAgwRgDoAVAC1gEMATASwDsBzAbgFgAobPfAVlY4GY8AJhwBhdgG92OGXkEA1CgBsArjAxIAPASwA+HAFkAFMAgQlOYABoxFOmBhKlFAC40IdYhADWMOjhcfPwBKaVkpNlkonBoAMxwTUMjolIwAdhw6GAB3HEVVdS0dXSMAIhcoNVLgvhSZMNSMrNz8tQ1tfD0jDUIAMQowQKgAT0IAZRcKKBcAORyjXptA3zoloLpgmoacAF9t7YAHKBoAN1cYLlxeowhgACsYQYD1pKiIurwMoyNROwcnVzuTwvZYhQgASQAzr97I5nG4PAAlGAARzUkJcMCoOAA/DhSmA/nCsaUcCB8bFlJCYKVarI9mwdkA==)
- [local function with state passing](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABADAAgwRgDoAVAC1gEMATASwDsBzAbgFgAobPfAVlY4GY8AJhwBhdgG92OGXkEA1CgBsArjAxIAPASwA+HAFkAFMAgQlOYABoxFOmBhKlFAC40IdYhADWMOjhcfPwBKaVkpNlkonBoAMxwTUMjolIwAdhw6GAB3HEVVdS0dXSMAIhcoNVLgvhSZMNSMrNz8tQ1tfD0jDUIAMQowQKgAT0IAZRcKKBcAORyjXptA3zoloLpgmobo7aiCJC5cXqMIYAArGEGA9aS6nAi7vYyjI1E7BydXd08b5ZDCACSAGc3vZHM43B4AEowACOaiBLhgVBwAH4cKUwO9wcjSjgQBjYsogTBSrU6gBfbZUtgUoA)

It'd be interesting to use BenchmarkDotNet to get some data on allocations for send/receive in various scenarios.

src/NetMQ/RoutingIdSocketExtensions.cs
```diff
@@ -269,22 +279,32 @@ public static ValueTask SendAsync([NotNull] this IRoutingIdSocket socket, uint r
...
             msg.InitEmpty();
 
-            socket.Receive(ref msg);
+            try
+            {
+                socket.Receive(ref msg, cancellationToken);
+            }
+            finally
+            {
+                msg.Close();
```
drewnoakes, line 298: Ditto, is it safe to close then use the `Msg` here?

src/NetMQ/ReceiveThreadSafeSocketExtensions.cs
somdoron: If I have to wrap both the socket and the token I'm not sure I have a benefit of using state instead of lambda.

src/NetMQ/Core/SocketBase.cs
somdoron: It is not throwing, I only throw from non-Try methods. The Try methods return false.

src/NetMQ/ReceiveThreadSafeSocketExtensions.cs
somdoron: good catch

src/NetMQ/ReceiveThreadSafeSocketExtensions.cs
somdoron: With extension method seems interesting:

https://sharplab.io/#gist:e7e76a5ece9d763149b8fb1eeed2587b

The question is, what is __ldftn is doing...

somdoron: seems to be allocation free:

https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.opcodes.ldftn?view=netcore-3.1

somdoron: solved it with private extension method.

drewnoakes: Actually casting `CancellationToken` as `object` will box, which allocates. I'm not sure there's a way around allocating here. I'd wondered about making the method `async ValueTask<>` but didn't test it.

drewnoakes: The main thing is that these methods don't allocate when they complete synchronously. If they need to yield, `ValueTask<>` will allocate anyway.

somdoron: Yes, you are right. I will simplify it back and merge it.

somdoron:
@drewnoakes next is IAsyncEnumerable, I wonder if you know of a convention to get the IAsyncEnumerable. One option is for the socket itself to implement it, however, that means we stick to one message type (byte[] probably). BlockingCollection has a method called `GetConsumingEnumerable`, I thought maybe to use that convention and to have:

1. GetConsumingStringAsyncEnumerable
2. GetConsumingBytesEnumerable

Anyway, not a big fan of it either. What do you think?

drewnoakes:
I haven't had much hands on experience with IAsyncEnumerable yet so don't have any strong opinions.

One naming idea to consider is maintaining prefix consistency to help discoverability via IntelliSense.

For example:

- `Receive`
- `ReceiveAsync`
- `ReceiveAsyncEnumerable`

PR #884, opened 2020-05-22
add AsyncEnumerable support to thread safe sockets

somdoron:


src/NetMQ/ReceiveThreadSafeSocketExtensions.cs
```diff
@@ -117,6 +119,31 @@ public static bool TryReceiveBytes(this IThreadSafeInSocket socket, [NotNullWhen
...
+        
+        /// <summary>
+        /// Provides a consuming IAsyncEnumerable for receiving messages from the socket.
+        /// </summary>
+        /// <param name="socket">The socket to receive from.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None"/>.</param>
+        /// <returns>An IAsyncEnumerable that receive and returns messages from the socket.</returns>
+        /// <exception cref="System.OperationCanceledException">The token has had cancellation requested.</exception>
+        public static async IAsyncEnumerable<byte[]> ReceiveBytesAsyncEnumerable(
+            [NotNull] this IThreadSafeInSocket socket,
```
drewnoakes, line 134:
```suggestion
            this IThreadSafeInSocket socket,
```

In a nullable context (i.e. where `#nullable enable` is used) then `[NotNull]`/`[DisallowNull]` is the default for un-annotated references.

Note that if you want to annotate that null is not a valid input (eg. on an unconstrainted generic reference) you must use `DisallowNull` (or `MaybeNull` for the inverse). `NotNull` in this case implies the reference is not null _after_ the call returns, and is usually used for `ref`/`out` params.

Ditto three more times below.

src/NetMQ.Tests/ClientServer.cs
```diff
@@ -71,7 +79,64 @@ public async void AsyncWithCancellationToken()
...
+            {
+                client.Send("0");
+            }
+            
+            // Send the end message to both of the threads
+            client.Send("1");
+            client.Send("1");
+
+            t1.Wait();
+            t2.Wait();
```
drewnoakes, line 137: nit: If this test could deadlock then awaiting these tasks with a timeout would be good.

src/NetMQ.Tests/ClientServer.cs
```diff
@@ -71,7 +79,64 @@ public async void AsyncWithCancellationToken()
...
+
+            for (int i = 0; i < 15000; i++)
+            {
+                client.Send("0");
+            }
+            
+            // Send the end message to both of the threads
+            client.Send("1");
+            client.Send("1");
+
```
drewnoakes, line 135: Could have a `totalCount` and assert it's 15,002 here to verify no dropped messages.

src/NetMQ.Tests/NetMQ.Tests.csproj
```diff
@@ -8,7 +8,7 @@
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
     <PackageTargetFallback Condition=" '$(TargetFramework)' == 'netcoreapp1.0' ">$(PackageTargetFallback);netcoreapp1.0;portable-net45+win8</PackageTargetFallback>
     <IsTestProject>true</IsTestProject>
-    <TargetFrameworks>netcoreapp2.1;net47</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;netcoreapp2.1;net47</TargetFrameworks>
```
drewnoakes, line 11: Consider dropping `netcoreapp2.1`.

somdoron: This is how I test netstandard2.0, so to drop netcoreapp2.1 would also mean to drop netstandard2.0

src/NetMQ/ReceiveThreadSafeSocketExtensions.cs
somdoron: thanks, fixed.

somdoron:
Added, thanks for the review, fixed it all.

PR #887, opened 2020-05-26
Annotate internal patterns, utils and some public API

drewnoakes:
Please especially review the few public API changes towards the end.

Note that `Assumes.NotNull` is `[Conditional("DEBUG")]` so has no runtime cost in release builds. I try to avoid using `!` where possible, and new code should be designed such that `Assumes.NotNull` isn't needed either. It's not reasonable to change the design on existing code to address all warnings, so adding these debug-only checks seems like a good tradeoff to me.

### DIFF
--- a/src/NetMQ/Core/Patterns/Server.cs
+++ b/src/NetMQ/Core/Patterns/Server.cs
@@ -67,8 +67,6 @@ namespace NetMQ.Core.Patterns
         /// <param name="icanhasall">not used</param>
         protected override void XAttachPipe(Pipe pipe, bool icanhasall)
         {
-            Assumes.NotNull(pipe);
-
             uint routingId = m_nextRoutingId++;
             if (routingId == 0)
                 routingId = m_nextRoutingId++; //  Never use Routing ID zero
--- a/src/NetMQ/Core/Patterns/Server.cs
+++ b/src/NetMQ/Core/Patterns/Server.cs
@@ -1,9 +1,6 @@
-#nullable disable
-
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using JetBrains.Annotations;
 using NetMQ.Core.Patterns.Utils;
 
 namespace NetMQ.Core.Patterns
@@ -20,13 +17,13 @@ namespace NetMQ.Core.Patterns
         /// </summary>
         private class Outpipe
         {
-            public Outpipe([NotNull] Pipe pipe, bool active)
+            public Outpipe(Pipe pipe, bool active)
             {
                 Pipe = pipe;
                 Active = active;
             }
 
-            [NotNull] public Pipe Pipe { get; }
+            public Pipe Pipe { get; }
 
             public bool Active;
         }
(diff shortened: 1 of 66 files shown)


drewnoakes:
@somdoron I cannot see why this test is failing. Is it known to be flakey? I see quite a few PRs merged with failing tests.

```
NetMQ.Tests.XPubSubTests.MultiplePublishers()
  X NetMQ.Tests.XPubSubTests.MultiplePublishers [509ms]
  Error Message:
   Assert.Equal() Failure
                                 ↓ (pos 23)
Expected: ···sage from subscriber
Actual:   ···sage from subscriber 2
                                 ↑ (pos 23)
  Stack Trace:
     at NetMQ.Tests.XPubSubTests.MultiplePublishers() in /home/runner/work/netmq/netmq/src/NetMQ.Tests/XPubSubTests.cs:line 279
```

somdoron:
yes, it is flakey...

src/NetMQ/Core/Patterns/Server.cs
```diff
@@ -70,7 +67,7 @@ public Server([NotNull] Ctx parent, int threadId, int socketId)
         /// <param name="icanhasall">not used</param>
         protected override void XAttachPipe(Pipe pipe, bool icanhasall)
         {
-            Debug.Assert(pipe != null);
+            Assumes.NotNull(pipe);
```
somdoron, line 70: I don't think this is needed anymore, not that SocketBase enabled nullable, probably for all sockets

src/NetMQ/Core/Patterns/Server.cs
```diff
@@ -186,7 +183,7 @@ protected override bool XRecv(ref Msg msg)
             if (!received)
                 return false;
 
-            Debug.Assert(pipe != null);
+            Assumes.NotNull(pipe);
```
somdoron, line 186: How smart is the compiler? Can he figure out that pipe cannot be null here? as you check the received boolean

src/NetMQ/Core/YPipe.cs
```diff
@@ -179,7 +178,7 @@ public bool CheckRead()
         /// Attempts to read an item from the pipe.
         /// </summary>
         /// <returns><c>true</c> if the read succeeded, otherwise <c>false</c>.</returns>
-        public bool TryRead(out T value)
+        public bool TryRead([MaybeNull] out T value)
```
somdoron, line 181: NotNullWhen?

somdoron:
LGTM, I think the compiler can figure out some of the assumes itself. You can merge :)

src/NetMQ/Core/Patterns/Server.cs
drewnoakes: Will remove.

src/NetMQ/Core/Patterns/Server.cs
drewnoakes: So actually the old code checked `pipe` against `null`, which showed the compiler that we thought the value might possibly be null, and so elevated the annotation on `pipe` from "not null" to "possibly null", which then led to warnings later on.

I converted a lot of `Debug.Assert(foo != null)` to `Assumes.NotNull(foo)` to get around these warnings. In this case (and the one above) it seems unlikely that `pipe` is ever null, so I'll remove both checks.

src/NetMQ/Core/YPipe.cs
drewnoakes: You're right. Updated.

Actually in this case it must be `MaybeNullWhen` because when `T` is unconstrained, it might be a value type for which `T?` actually means `Nullable<T>`. In C# 8 you cannot annotate an unconstrained generic type with `?`.

In C# 9 the team is investigating allowing `T??` as a syntax for exactly this situation.

PR #896, opened 2020-05-27
Further nullability annotations

drewnoakes:
Covers most code. Only five files have `#nullable disable` now.

### DIFF
--- a/src/NetMQ.Tests/EventDelegatorTests.cs
+++ b/src/NetMQ.Tests/EventDelegatorTests.cs
@@ -1,6 +1,4 @@
-﻿#nullable disable
-
-using System;
+﻿using System;
 using Xunit;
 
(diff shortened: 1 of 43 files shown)


src/NetMQ.Tests/EventDelegatorTests.cs
```diff
@@ -40,6 +38,7 @@ void DelegatorHandler(object sender, Args<double> args)
             delegator.Event += DelegatorHandler;
 
             Assert.NotNull(Source);
+            Assumes.NotNull(Source);
```
drewnoakes, line 41: xUnit won't be annotated until v3 is released, so there are a few suppressions following `Asset.NotNull` in tests unfortunately.

https://github.com/xunit/xunit/pull/2044

PR #914, opened 2020-07-07
Fixed possible NullReferenceException during shutdown of failed conne…

dxdjgl:
…ction.


src/NetMQ/Core/Mechanisms/CurveMechanismBase.cs
```diff
@@ -45,7 +45,7 @@ public override void Dispose()
         {
             Assumes.NotNull(m_box);
```
drewnoakes, line 46: Should remove this line too.

dxdjgl: I agree, line has been removed and I am creating a new pull request.

PR #917, opened 2020-07-09
Provision to have override long running poller 

AnandKR84:
Marking the Poller thread a background thread since the process hosting this poller is not shutting down.

### DIFF
--- a/src/NetMQ/NetMQPoller.cs
+++ b/src/NetMQ/NetMQPoller.cs
@@ -430,7 +430,10 @@ namespace NetMQ
             if (IsRunning)
                 throw new InvalidOperationException("NetMQPoller is already running");
 
-            var thread = new Thread(Run) { Name = threadName };
+            var thread = new Thread(Run) { 
+                 Name = threadName,                              
+		         IsBackground = true
+            };
             thread.Start();
 
             m_switch.WaitForOn();
--- a/src/NetMQ/NetMQPoller.cs
+++ b/src/NetMQ/NetMQPoller.cs
@@ -17,119 +17,119 @@ using Switch = NetMQ.Core.Utils.Switch;
 
 namespace NetMQ
 {
-    /// <summary>
-    /// Enable polling on multiple NetMQSockets
-    /// </summary>
-    public sealed class NetMQPoller :
+	/// <summary>
+	/// Enable polling on multiple NetMQSockets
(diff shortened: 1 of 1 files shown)


drewnoakes:
This will potentially break applications that currently depend upon the poller thread being a foreground thread.

drewnoakes:
Are you not able to dispose the poller in your application to have graceful exit?

AnandKR84:
> Are you not able to dispose the poller in your application to have graceful exit?

Yes, that is possible, but in cases where the poller is in a singleton class, we are finding this issue. (I know that singleton is an anti-pattern, but in our case, it is needed to be a singleton class.)

drewnoakes:
Can you dispose your singleton as your `Main` method exits?

drewnoakes:
This is a source-compatible change, but not a binary-compatible one.

@somdoron what's the bar we set for these kinds of changes? If someone updates the library without recompiling (e.g. imported as a transitive dependency) they'll get `MethodMissingException` at runtime. Do we need an overload, or perhaps a new property to control this?

somdoron:
I think the solution is pretty simple, just to add another overload and it will binary compatible?

drewnoakes:
@AnandKR84 can you update this PR to overload the method?

somdoron:
@AnandKR84 any reason for closing the PR?

AnandKR84:
> @AnandKR84 any reason for closing the PR?

It was by mistake that was done. There was some formatting changes since i did it in a new editor.

dxdjgl:
I know it is a bit of nitpicking, but I dont like the name longRunningPoller, which also causes a negation inside RunAsync method.
I think I would prefer somehting like IsBackgroundThread, otherwise I am okay with the code now.

dxdjgl:
@somdoron I think the code is fine now do you agree?

dxdjgl:
Since the negation has gone, the the value passed in public void RunAsync(string threadName) must be changed as well.

@somdoron when this change has been applied, the I think it should be okay to be pulled?

AnandKR84:
> Since the negation has gone, the the value passed in public void RunAsync(string threadName) must be changed as well.
> 
> @somdoron when this change has been applied, the I think it should be okay to be pulled?

Thanks @dxdjgl for pointing it out.

PR #920, opened 2020-07-12
Null termination is not a readable character, and encoded string shou…

dxdjgl:
…ld only be 40 characters long.

The null termination, would be interpreted as an space, causing the decoding to fail.


dxdjgl:
@somdoron due to this error, the Z85 encoding is not that useful, how can I initiate a new release build?

PR #922, opened 2020-07-17
Update message.md

Andor233:
Add code block ending，
The origin one can display correctly in github but wrong in readthedocs


drewnoakes:
Thank you!

PR #924, opened 2020-07-24
Updated NaCl.Net to the 0.1.13 version to have increased performance,…

dxdjgl:
… with encrypted connections.


dxdjgl:
@somdoron thank you very much for the new build of NaCl.net, would you initiate a new release of NetMQ after this pull request?

PR #935, opened 2020-09-11
Fix TrySendFrame if buffer is bigger than length

Amberg:
"TrySendFrame" thrown a System.ArgumentException if buffer was bigger than length


Amberg:
@somdoron Would you initiate a new release of NetMQ after this pull request?
Thank you

somdoron:
done

PR #938, opened 2020-09-24
NetMQMonitor Can Throw "Unnecessarily"

mmillerbe:
Previous code casted to the expected type directly, which succeeded in the case `monitorEvent.Arg` was null.  The current `is` check will be false for null values, causing an exception to be thrown which the attached Poller doesn't handle.


mmillerbe:
I pushed up a slight refactoring here that added some tests so that the code coverage wouldn't decrease.  Is that more acceptable?  I'm happy to make whatever changes you'd like if there's some other way you think this should be handled.  (To suppress warnings, I needed to add a nullable indicator on a property in a public-facing API.  I think it's "correct" in that the socket can be null, but I'm not sure if that's desirable from your perspective.)

Thanks!
Mike

jgasperlin:
Why is this not merged yet ?

drewnoakes:
Apologies for the delay here. Thanks for the contribution.

mmillerbe:
No worries. Thanks, @drewnoakes!

jgasperlin:
when is new release planned to include this change ?

urosek:
Is there any update on this?

drewnoakes:
4.0.1.9 has been published to NuGet and includes this change:

https://www.nuget.org/packages/NetMQ/4.0.1.9

PR #943, opened 2020-10-12
Default encoding should be UTF-8 as per standard (ref IETF RFC 7159)

theonlynexus:
In page 9 of the [RFC](https://tools.ietf.org/html/rfc7159), there is the following paragraph:

"8.1.  Character Encoding
   JSON text SHALL be encoded in UTF-8, UTF-16, or UTF-32.  The default
   encoding is UTF-8, and JSON texts that are encoded in UTF-8 are
   interoperable[...]"


theonlynexus:
No plans to merge?

src/NetMQ/NetMQFrame.cs
```diff
@@ -54,7 +54,7 @@ public NetMQFrame(byte[] buffer, int length)
         /// </summary>
         /// <param name="message">a string containing the message-data of the frame</param>
         public NetMQFrame(string message)
```
drewnoakes, line 56: The doc comment on this member still mentions ASCII.

drewnoakes (commented):
Could you provide some more context on this change please? The RFC you linked is for JSON, which has no relevance here. You would need to check the ZeroMQ specification to know whether this is a valid change to the library. Can you find such a reference and share it please?

src/NetMQ/NetMQFrame.cs
theonlynexus: First of all, thanks for replying.

You are right in pointing out that I linked to the JSON RFC, not ZMQ's docs: couldn't tell you why, right now.
At the time I submitted my PR, I had tracked down my problems, but I can't find my notes anymore.

Also, as you point out, I don't see anything in the ZMQ specification that indicates that UTF-8 or JSON should be used, in fact the contrary is correct: "On the wire, ZeroMQ messages are blobs of any size from zero upwards that fit in memory. You do your own serialization using protocol buffers, msgpack, JSON, or whatever else your applications need to speak." (from [the messages docs](https://zeromq.org/messages/))

I'll try to find why I thought that was the correct point to set the encoding, but I might just have been wrong the whole time. In that case I'll just close the PR.

Again, thanks for the reply!

theonlynexus: I can't seem to find anything regarding my old thoughts for opening the PR, and looking at it now it seems rather nonsensical. I can only assume that I was sleep deprived.

Closing it.

Thanks for taking the time to give it a look.

theonlynexus:
As mentioned in the conversation thread, I am closing the PR because in retrospect it seems rather non-sensical. Not sure what was going through my mind when I opened it.

PR #960, opened 2020-12-31
Poller tests fix

### LINKED ISSUE #959: Multiple failing unit tests
### Environment

    NetMQ Version:    latest master branch
    Operating System: Win10
    .NET Version:     

### Expected behaviour
Unit tests run and pass

### Actual behaviour

Many tests fail

Run unit tests.

Is this expected on the current master branch?

jasells:
This fixes #959 for running unit tests in Debug.

### DIFF
--- a/src/NetMQ.Tests/NetMQPollerTest.cs
+++ b/src/NetMQ.Tests/NetMQPollerTest.cs
@@ -51,8 +51,8 @@ namespace NetMQ.Tests
             }
         }
 
-        [Fact]
-        public void Monitoring()
+        [Fact(Timeout=5000)]
+        public async void Monitoring()
         {
             var listeningEvent = new ManualResetEvent(false);
             var acceptedEvent = new ManualResetEvent(false);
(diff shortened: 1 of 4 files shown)


jasells:
All tests run locally, not sure what the problem is

valeriob:
I encountered the same problem, is this fix ready ?

dakelley95:
Encountered same issue as well.  Would love this fix.

src/NetMQ.Tests/NetMQPollerTest.cs
```diff
@@ -81,11 +81,27 @@ public void Monitoring()
                 req.Connect("tcp://127.0.0.1:" + port);
                 req.SendFrame("a");
 
-                rep.SkipFrame();
+                // keep the test from blocking when something goes wrong.
+                var timeout = new CancellationTokenSource(5000).Token;
+                Exception e = null;
```
drewnoakes, line 86:
```suggestion
                Exception? e = null;
```

src/NetMQ/Core/Transports/ByteArraySegment.cs
```diff
@@ -57,7 +57,7 @@ public ByteArraySegment(ByteArraySegment otherSegment, int offset)
         /// <summary>
         /// Get the number of bytes within the buffer that is past the Offset (ie, buffer-length minus offset).
         /// </summary>
-        public int Size => m_innerBuffer.Length - Offset;
+        public int Size => m_innerBuffer == null ? 0 : m_innerBuffer.Length - Offset;
```
drewnoakes, line 60: I'm not sure about this change. Code shouldn't be using an initialised version of this struct. The buffer should always be non-null. If it's null here, then there's a bug somewhere else. That is the bug that should probably be identified and fixed. At least that's my working assumption.

PR #966, opened 2021-02-11
Fix cancellation in ReceiveFrameBytesAsync

### LINKED ISSUE #965: ReceiveFrameBytesAsync does not deregister from cancellationtoken, leading to a range of inner exceptions (also creates a memory leak)
### Environment

    NetMQ Version:    4.0.1.6
    Operating System: Windows 10 20H2
    .NET Version:     5.0

### Expected behaviour

When terminating the service using Ctrl+C on the console I expect it to terminate without dozens or hundreds of unhandled exceptions.

### Actual behaviour

The longer you let it run, the more inner exceptions are received, potentially thousands or more.

Receiving thousands of inner exceptions of the form:
 ---> (Inner Exception #34) System.InvalidOperationException: An attempt was made to transition a task to a final state when it had already completed.
   at System.Threading.Tasks.TaskCompletionSource`1.SetCanceled(CancellationToken cancellationToken)
   at System.Threading.Tasks.TaskCompletionSource`1.SetCanceled()
   at NetMQ.AsyncReceiveExtensions.<>c__DisplayClass3_0.<ReceiveFrameBytesAsync>b__0()
   at System.Threading.CancellationToken.<>c.<.cctor>b__26_0(Object obj)
   at System.Threading.CancellationTokenSource.CallbackNode.<>c.<ExecuteCallback>b__9_0(Object s)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.CancellationTokenSource.CallbackNode.ExecuteCallback()
   at System.Threading.CancellationTokenSource.ExecuteCallbackHandlers(Boolean throwOnFirstException)<---
<---

I created an XPub/XSub proxy and a couple of services using a .NET 5.0 Worker Service project. The project is very simple and should show only the issue. The services only pass messages between each other following a delay of 100ms.

1. Clone this project: https://github.com/abeham/netmq-proxy
2. Open a terminal within this repository root folder
3. dotnet run -p Middleware
4. Hit Ctrl+C after some seconds

You experience a range of exceptions being printed to the console, the longer you wait before aborting with Ctrl+C, the more exceptions will be produced.

### The problem

The method registeres a handler for the internally used TaskCompletionSource on the cancellation token. If that token is however injected from outside, the task completion source is still alive if the ReceiveFrameBytesAsync method terminates, because the source is never deregistered.

jkarder:
This PR updates ReceiveFrameBytesAsync to detach competing listeners before finishing the TaskCompletionSource and call `TrySetResult` and `TrySetCanceled` instead of `SetResult` and `SetCanceled`.

Fixes #965.

### DIFF
--- a/src/NetMQ/AsyncReceiveExtensions.cs
+++ b/src/NetMQ/AsyncReceiveExtensions.cs
@@ -85,7 +85,7 @@ namespace NetMQ
             }
 
             TaskCompletionSource<(byte[], bool)> source = new TaskCompletionSource<(byte[], bool)>();
-            cancellationToken.Register(() => source.SetCanceled());
+            var registration = cancellationToken.Register(PropagateCancel);
 
             void Listener(object sender, NetMQSocketEventArgs args)
             {
@@ -96,8 +96,16 @@ namespace NetMQ
                     msg.Close();
 
                     socket.ReceiveReady -=  Listener;
-                    source.SetResult((data, more));
+                    registration.Dispose();
+                    source.TrySetResult((data, more));
                 }
+            }
+            
+            void PropagateCancel()
+            {
+                socket.ReceiveReady -= Listener;
+                registration.Dispose();
+                source.TrySetCanceled();
             }
 
             socket.ReceiveReady += Listener;
--- a/src/NetMQ/AsyncReceiveExtensions.cs
+++ b/src/NetMQ/AsyncReceiveExtensions.cs
@@ -20,7 +20,7 @@ namespace NetMQ
         static Task<bool> s_trueTask = Task.FromResult(true);
         static Task<bool> s_falseTask = Task.FromResult(false);
 
-        #region Receiving frames as a multipart message
+#region Receiving frames as a multipart message
 
         /// <summary>
         /// Receive a single frame from <paramref name="socket"/>, asynchronously.
@@ -52,9 +52,9 @@ namespace NetMQ
             return message;
         }
 
-        #endregion
+#endregion
 
-        #region Receiving a frame as a byte array
+#region Receiving a frame as a byte array
 
         /// <summary>
         /// Receive a single frame from <paramref name="socket"/>, asynchronously.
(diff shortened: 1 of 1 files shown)


drewnoakes:
The same treatment is needed in `ReceiveFrameStringAsync`. I don't know why `SkipFrameAsync` doesn't support cancellation, but you could add it there too as the pattern is similar.

Also, it's worth checking `token.CanBeCanceled`, as `default(CancellationToken)` will never be cancelled and so there's no point registering. For many users, no cancellation will be required, so we can avoid the callback registration overhead completely.

drewnoakes (approved):

drewnoakes (changes requested):
Just noticed the test failures. Please fix those up and we'll get this merged.

@somdoron does this look good to you? There is an API change here.

somdoron (approved):

somdoron:
LGTM

abeham:
The tests succeed locally on my Windows 10 installation for netcoreapp2.1, netcoreapp3.1, and net47. Maybe the build server had some hickups, because the failing test is a timeout.

jkarder:
Thanks for your approvals. The tests succeed here as well (Windows 10.0.19041).

netcoreapp2.1:
```
Test Run Successful.
Total tests: 256
     Passed: 240
    Skipped: 16
 Total time: 37.3524 Seconds
```

netcoreapp3.1:
```
Test Run Successful.
Total tests: 258
     Passed: 242
    Skipped: 16
 Total time: 38.2665 Seconds
```

net47:
```
Test Run Successful.
Total tests: 259
     Passed: 243
    Skipped: 16
 Total time: 39.0455 Seconds
```

BTW, would you like me to revert the whitespace changes I accidentally committed in d297a27?

drewnoakes:
Re-running jobs to see if that helps.

> BTW, would you like me to revert the whitespace changes I accidentally committed in d297a27?

Sure, but only if it's convenient.

drewnoakes:
Yep, just needed a kick. I'll merge as-is. Thanks very much!

Issue #978, opened 2021-05-27
Msg does not expose Memory<byte> of the data.


Svisstack:
### Environment

    NetMQ Version:    4.0.16

### Expected behaviour

Msg will expose Span<byte> Slice() and Memory<byte> SliceAsMemory();

### Actual behaviour

Data is marked as Obsolete and only Span<byte> Slice() is exposed.

### Steps to reproduce the behaviour

1. Have need to have ReadOnlyMemory<byte> representing data of the Message.
2. Want to have a code that will not need a change on the next major release by being dependent on Obsolete.

I was wondering if this is on purpose or that function could be merged into this codebase.

drewnoakes:
> I was wondering if this is on purpose or that function could be merged into this codebase.

I'm not aware of any reason not to do this. Would you like to make a pull request?

PR #982, opened 2021-06-29
Issue #978 expose internal buffer as memory<byte> #1

Svisstack:
zeromq#978

### DIFF
--- a/src/NetMQ/Msg.cs
+++ b/src/NetMQ/Msg.cs
@@ -265,6 +265,18 @@ namespace NetMQ
             
             return new Span<byte>(m_data, m_offset, Size);  
         }
+        
+        /// <summary>
+        /// Return the internal buffer as Memory
+        /// </summary>
+        /// <returns>The memory</returns>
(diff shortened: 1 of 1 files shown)


src/NetMQ/Msg.cs
```diff
@@ -265,6 +265,18 @@ public Span<byte> Slice()
             
             return new Span<byte>(m_data, m_offset, Size);  
         }
+        
+        /// <summary>
+        /// Return the internal buffer as Memory
+        /// </summary>
+        /// <returns>The memory</returns>
+        public Memory<byte> SliceAsMemory()
```
drewnoakes, line 273: It's a shame We didn't use `AsSpan` instead of `Slice`, so this could be `AsMemory` too.

drewnoakes (approved):

src/NetMQ/Msg.cs
Svisstack: Yes, it's ok or you want to rename it?

Issue #225, opened 2015-02-06
3.3.0.12-rc1 OverflowException


SemtexCharlie:
Got the following stacktrace from a process that errored using the Nuget sourced 3.3.0.12-rc1. 

```
System.OverflowException: Arithmetic operation resulted in an overflow.
   at NetMQ.GCBufferPool.Take(Int32 size)
   at NetMQ.Msg.InitPool(Int32 size)
   at NetMQ.zmq.Transports.V2Decoder.EightByteSizeReady()
   at NetMQ.zmq.Transports.DecoderBase.ProcessBuffer(ByteArraySegment data, Int32 size)
   at NetMQ.zmq.Transports.StreamEngine.ProcessInput()
   at NetMQ.zmq.Transports.StreamEngine.FeedAction(Action action, SocketError socketError, Int32 bytesTransferred)
   at NetMQ.zmq.Utils.Proactor.Loop()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()
```
I don't have a capture of the message that caused it, but I can say that it was the **only socket** in the context/executing process (regular PUB socket) and it was open/established well ahead of this exception - suggesting it was the SentMessage() call I think.

Here's the simple send loop, no polling etc... (qItem type is `NetMQMessage`)

```csharp
foreach (var qItem in OutboundQueue.GetConsumingEnumerable(HaltFlag.Token))
{
  try
  {
    // Log Debug sending msg text removed
    Socket.SendMessage(qItem);
    SentMessageCount++;
  }
  catch (Exception ex)
  {
    // Log Error statement removed
  }
}
```

Haven't been able to reproduce it yet but will keep trying with more more granular debug statements... 

Should I fall back to 3.3.0.11? Did it have the proactor too?

SemtexCharlie:
Update:

- Does happen with latest github code
- Doesn't happen with 3.3.0.11

More about causing this to happen...

[...]

PR #1009, opened 2022-03-24
Implement similar fix to issue #225 (fixed in #388) for V1Decoder

rheatley-pervasid:
We have seen a customer suffer a crash with the following stack trace

```
Exception Info: System.OverflowException
   at NetMQ.GCBufferPool.Take(Int32)
   at NetMQ.Msg.InitPool(Int32)
   at NetMQ.Core.Transports.V1Decoder.EightByteSizeReady()
   at NetMQ.Core.Transports.V1Decoder.Next()
   at NetMQ.Core.Transports.DecoderBase.Decode(NetMQ.Core.Transports.ByteArraySegment, Int32, Int32 ByRef)
   at NetMQ.Core.Transports.StreamEngine.ProcessInput()
   at NetMQ.Core.Transports.StreamEngine.Handle(Action, System.Net.Sockets.SocketError, Int32)
   at NetMQ.Core.Transports.StreamEngine.FeedAction(Action, System.Net.Sockets.SocketError, Int32)
   at NetMQ.Core.Transports.StreamEngine.InCompleted(System.Net.Sockets.SocketError, Int32)
   at NetMQ.Core.IOObject.InCompleted(System.Net.Sockets.SocketError, Int32)
   at System.Threading.ThreadHelper.ThreadStart_Context(System.Object)
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
```

This was on version 4.0.1.6.
The crash is in our application which has a RouterSocket bound to "tcp://127.0.0.1:9202"

I tracked down issues #225 and #388 and believe this is the same problem.
I have so far failed to reproduce the issue using NetMQ or python zmq, it isn't clear to me how to send a large enough message.

I can provoke the crash with some contrived python code
```
import socket
s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.connect(("127.0.0.1", 9202))
s.send(b'\xFF' * 9 + b'\x00' + b'\x42' * 1000000)
```

Taking the fix from #388 and applying to V1Decoder.cs prevents the above python code crashing NetMQ.
There is a comment in that PR - "Other decoders/encoders are worth checking out too I guess." - I assume this never happened?

Let me know your thoughts

### DIFF
--- a/src/NetMQ/Core/Transports/V1Decoder.cs
+++ b/src/NetMQ/Core/Transports/V1Decoder.cs
@@ -121,18 +121,19 @@ namespace NetMQ.Core.Transports
 
             //  8-byte payload length is read. Allocate the buffer
             // for message body and read the message data into it.
-            long payloadLength = m_tmpbuf.GetLong(Endian, 0);
+            ulong payloadLength = m_tmpbuf.GetUnsignedLong(Endian, 0);
 
             // There has to be at least one byte (the flags) in the message).
             if (payloadLength == 0)
                 return DecodeResult.Error;
 
             // Message size must not exceed the maximum allowed size.
-            if (m_maxMessageSize >= 0 && payloadLength - 1 > m_maxMessageSize)
+            if (m_maxMessageSize >= 0 && payloadLength - 1 > (ulong)m_maxMessageSize)
                 return DecodeResult.Error;
 
+            // TODO: move this constant to a good place (0x7FFFFFC7)
             // Message size must fit within range of size_t data type.
-            if (payloadLength - 1 > int.MaxValue)
+            if (payloadLength - 1 > 0x7FFFFFC7)
                 return DecodeResult.Error;
 
             int msgSize = (int)(payloadLength - 1);


rheatley-pervasid:
@drewnoakes anything I can do to help progress this?

drewnoakes:
Apologies for dropping the ball on this.

I'm on my phone and don't have the spec handy. If the type is unsigned then the fix here would be more comprehensive.

drewnoakes:
> Oh, looks like it got fixed differently in #1030 Assume you are happy to lose support for messages >2GB and I'll close it?

So looking at this PR again (sorry again for the delay) I am wondering if this is any different in practice to the previous change that checked for negative values.

While this will correctly pull a payload length greater than 2GB, it then goes on to validate that it's no longer than `0x7FFFFFC7`, which is actually _less_ than `int.MaxValue`. I don't know where that constant comes from, and it seems more restrictive than what's on the main branch now.

PR #1017, opened 2022-05-08
Remove System.ServiceModel.Primitives dependency.

shanecelis:
In an effort to use NetMQ with Unity3D, one problem was trying to 
satisfy the dependencies for `System.ServiceModel.Primitives`. It
required also providing `System.ServiceModel.{Http,Duplex,NetTcp}` and
perhaps more before I gave up. These dependencies were only noticed
while Unity3D built the project. When using NetMQ in a regular dotnet
project, I've had no problems with the `System.ServiceModel.Primitives`
dependencies.

I added an alternative implementation of `IBufferPool` using the
`ArrayPool` from `System.Buffers` but it is unused. I instrumented the
build system to have a "Flavor" that by default is set to "Standard",
which changes nothing from before.

For my usecase, however, I set the flavor to "Minimal".

    $ dotnet build -p:Flavor=Minimal

This excludes "System.ServiceModel.Primitives" and defines a
preprocessor flag of `FLAVOR_MINIMAL` that excludes the
BufferManagerBufferPool class and uses the ArrayBufferPool instead.

### DIFF
--- a/src/NetMQ.Tests/BufferPoolTests.cs
+++ b/src/NetMQ.Tests/BufferPoolTests.cs
@@ -1,0 +1,64 @@
+using System;
+using Xunit;
+
+namespace NetMQ.Tests
+{
+    public abstract class BufferPoolTests
+    {
+        protected int maxBufferSize = 2;
+        protected long maxBufferPoolSize = 100L;
+        protected IBufferPool pool;
+        
+        [Theory]
+        [InlineData(500)]
+        [InlineData(50000)]
+        [InlineData(5000000)]
+        [InlineData(500000000)]
+        public void Rent(int size)
+        {
+            var array = pool.Take(size);
+            Assert.Equal(size, array.Length);
+        }
+        
+        // This never happens.
+        // [Fact]
+        // public void RentTooBigBuffer()
+        // {
+        //     Assert.ThrowsAny<Exception>(() => pool.Take(900000000));
+        // }
+
+        [Fact]
+        public void Return() {
+          var array = pool.Take(10);
+          pool.Return(array);
+        }
+
+        [Fact]
+        public void ReturnUnknown() {
+          var array = new byte[100];
+          Assert.Throws<ArgumentException>(() => pool.Return(array));
+        }
+    }
+
+    public class ArrayBufferPoolTests : BufferPoolTests
+    {
+        public ArrayBufferPoolTests()
+        {
+            /* The sizes were chosen to be very small. It's not clear they
+             * constitute any bounds on the array sizes that can be
+             * requested. */
+            pool = new ArrayBufferPool(maxBufferPoolSize, maxBufferSize);
+        }
+    }
+
+    public class BufferManagerBufferPoolTests : BufferPoolTests
+    {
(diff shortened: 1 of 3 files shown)


drewnoakes:
Is there a reason not to outright replace the prior implementation? Losing the dependency on a WCF library seems sensible.

shanecelis:
There's no reason that I know. I did it this way merely to make the least invasive change. 

Technically using `ArrayPool<T>` introduces a dependency on `System.Buffers` but that's already a dependency that's required by `System.Memory` which `NaCl.Net` requires. Let me know if you want to drop the dependency all together and get rid of the build flavor stuff.

drewnoakes:
I'd vote for just replacing the WCF implementation outright, and removing the dependency on that package.

The implementation is here for reference: https://github.com/CoreWCF/CoreWCF/blob/f7113a7f2537ecd4d5ca397fbebbbb97aad241a1/src/CoreWCF.Primitives/src/CoreWCF/Runtime/InternalBufferManager.cs

@somdoron, any thoughts on this?

shanecelis:
In response to the feedback and code coverage report given above, I added tests for both pool implementations, and based on those tests I added a fix to make `ArrayBufferPool`'s behavior conform to `BufferManagerBufferPool`'s. 

I removed the `Flavor` build parameter, which simplifies the build back to what it was. 

I removed the reference to `System.ServiceModel.Primitives` and added an explicit reference to `System.Buffers` even though it's implicitly required by `NaCl.Net`

I did add a USE_SERVICE_MODEL define so that we can exercise tests against both implementations for the interim.

On macOS, I couldn't run the full test suite. (Something hangs but I didn't investigate it.) I did exercise ArrayBufferPoolTests and BufferManagerBufferPoolTests that I wrote and they pass.

From the testing I did notice one peculiarity. No matter how small the pool size parameters are, one can rent an array much larger than the max array size. This is true with the original implementation and the new implementation preserves this behavior but it seems odd. I would expect an exception to be thrown if I requested an array size that was bigger than my pool's max array size. But that doesn't happen here currently. See `BufferPoolTests.cs` to see for yourselves.

shanecelis:
Is this failure genuinely a result of my changes? I'm having trouble running the test suite on macOS. I'll try on Windows another time.

dxdjgl:
Please be aware that the ArrayBufferPool behaves very differently than the existing BufferManagerBufferPool, very different values needs to be passed for the construction.

AlexandreArpin:
Bumping this, I'd be more than happy to help in this area.

Is it confirmed that WCF's BufferManager should be replaced by ArrayPool? I'm cautious about the implications of changing behavior for existing users, I'm not familiar enough with the typical 'renting/allocation' scenarios and/or the difference/nuance between WCF's BufferManager and ArrayPool. There's also the question of how to honor/convert the parameters passed to `BufferPool.SetBufferManagerBufferPool`

Alternatively, we could "vendor" the WCF's BufferManager class to remove the dependency on the library, keeps backward compatibility and avoids all headaches of backward compatibility/performance/behavior regressions? ArrayPool could always be added as a new options to NetMQ's `BufferPool` if that is an option

PR #1050, opened 2023-03-12
Fix issue where message is assumed to contain correct length byte (leading to process exit)

### LINKED ISSUE #1049: ArgumentOutOfRangeException while processing handshake causes process exit
### Environment

    NetMQ Version:    4.0.1.10
    Operating System: Windows 11 x64
    .NET Version:     7.0.103

### Expected behaviour

Ability for NetMQ to handle/catch errors, and/or provide an error handling mechanism that users can plug in to. It is not nice to have code in your process that can cause it to abrubtly exit.

### Actual behaviour

The process terminates.

```
2023-03-06 00:44:24.880 [FTL] Host Caught unhandled app "ArgumentOutOfRangeException" in "api", "UNABLE TO PREVENT PROCESS EXIT!"
System.ArgumentOutOfRangeException: Index and count must refer to a location within the buffer. (Parameter 'bytes')
   at System.Text.ASCIIEncoding.GetString(Byte[] bytes, Int32 byteIndex, Int32 byteCount)
   at NetMQ.Msg.GetString(Encoding encoding, Int32 offset, Int32 count)
   at NetMQ.Core.Mechanisms.Mechanism.IsCommand(String command, Msg& msg)
   at NetMQ.Core.Mechanisms.NullMechanism.ProcessHandshakeCommand(Msg& msg)
   at NetMQ.Core.Transports.StreamEngine.ProcessHandshakeCommand(Msg& msg)
   at NetMQ.Core.Transports.StreamEngine.ProcessInput()
   at NetMQ.Core.Transports.StreamEngine.Handle(Action action, SocketError socketError, Int32 bytesTransferred)
   at NetMQ.Core.Transports.StreamEngine.FeedAction(Action action, SocketError socketError, Int32 bytesTransferred)
   at NetMQ.Core.Transports.StreamEngine.InCompleted(SocketError socketError, Int32 bytesTransferred)
   at NetMQ.Core.IOObject.InCompleted(SocketError socketError, Int32 bytesTransferred)
```

I'm not sure, but someone (this is an API server where each client has their own set of DEALER/XPUB sockets to connect to) is connecting to the port and sending data that NetMQ interprets as a handshake command. The code to determine whether the Msg is a command seems to make invalid assumptions about the Data in the Msg and fails when it tries to get the command name. 

Because the entire call-chain is without exception handler, the exception is only caught by the `AppDomain.CurrentDomain.UnhandledException += AppOnUnobservedException;` handler, after which the process terminates.

mnmr:
Fixes #1049

The original code only verified the length after calling GetString, but never gets there if the subsequent call to encoding.GetString fails with an ArgumentOutOfRangeException. This change moves the verification to before the call to GetString.

The second commit modifies the Proactor.Loop method so that it does not trigger a process exit due to an unhandled exception when an error occurs. To this end, it no longer throws an ArgumentOutOfRangeException if an invalid completion status is encountered (incidentally, this is the same error thrown by the GetString method for the earlier bug), and it catches all exceptions to allow the loop to continue in case of a failure.

### DIFF
--- a/src/NetMQ.Tests/MechanismTests.cs
+++ b/src/NetMQ.Tests/MechanismTests.cs
@@ -1,0 +1,55 @@
+﻿using System.Text;
+using NetMQ.Core.Mechanisms;
+using Xunit;
+
+namespace NetMQ.Tests
+{
+    public class MechanismTests
+    {
+        private Msg CreateMsg(string data, int lengthDiff)
+        {
+            Assert.NotNull(data);
+            Assert.True(data.Length > 1);
+            var length = data.Length + lengthDiff;
+            Assert.True(length > 0);
+            Assert.True(length < byte.MaxValue - 1);
+            
+            var msg = new Msg();
+            msg.InitGC(new byte[data.Length + 1], 0, data.Length + 1);
+            msg.SetFlags(MsgFlags.Command);
+            msg.Put((byte)length);
+            msg.Put(Encoding.ASCII, data, 1);
+            return msg;
+        }
+
+        [Fact]
+        public void IsCommandShouldReturnTrueForValidCommand()
+        {
+            var mechanism = new NullMechanism(null, null);
+            var msg = CreateMsg("READY", 0);
+            Assert.True(mechanism.IsCommand("READY", ref msg));
+        }
+
+        [Fact]
+        public void IsCommandShouldReturnFalseForInvalidCommand()
+        {
+            var mechanism = new NullMechanism(null, null);
+            var msg = CreateMsg("READY", -1);
+            Assert.False(mechanism.IsCommand("READY", ref msg));
+            msg = CreateMsg("READY", 1);
+            Assert.False(mechanism.IsCommand("READY", ref msg));
+            // this test case would fail due to an exception being throw (in 4.0.1.10 and prior)
+            msg = CreateMsg("READY", 2);
+            Assert.False(mechanism.IsCommand("READY", ref msg));
+        }
+
+        // this test was used to validate the behavior prior to changing the validation logic in Mechanism.IsCommand
+        // [Fact]
+        // public void IsCommandShouldThrowWhenLengthByteExceedsSize()
+        // {
+        //     var mechanism = new NullMechanism(null, null);
+        //     var msg = CreateMsg("READY", 2);
+        //     Assert.Throws<ArgumentOutOfRangeException>(() => mechanism.IsCommand("READY", ref msg));
+        // }        
+    }
+}
--- a/src/NetMQ.Tests/MechanismTests.cs
+++ b/src/NetMQ.Tests/MechanismTests.cs
@@ -25,7 +25,7 @@ namespace NetMQ.Tests
         [Fact]
         public void IsCommandShouldReturnTrueForValidCommand()
         {
-            var mechanism = new NullMechanism(null, null);
+            var mechanism = new NullMechanism(null!, null!);
             var msg = CreateMsg("READY", 0);
             Assert.True(mechanism.IsCommand("READY", ref msg));
         }
@@ -33,7 +33,7 @@ namespace NetMQ.Tests
         [Fact]
         public void IsCommandShouldReturnFalseForInvalidCommand()
         {
-            var mechanism = new NullMechanism(null, null);
+            var mechanism = new NullMechanism(null!, null!);
             var msg = CreateMsg("READY", -1);
             Assert.False(mechanism.IsCommand("READY", ref msg));
             msg = CreateMsg("READY", 1);
--- a/src/NetMQ/Core/Mechanisms/Mechanism.cs
+++ b/src/NetMQ/Core/Mechanisms/Mechanism.cs
@@ -348,7 +348,7 @@ namespace NetMQ.Core.Mechanisms
             return true;
         }
         
-        protected bool IsCommand(string command, ref Msg msg)
+        internal protected bool IsCommand(string command, ref Msg msg)
         {
             if (msg.Size >= command.Length + 1 && msg[0] == command.Length)
             {
--- a/src/NetMQ/Core/Mechanisms/Mechanism.cs
+++ b/src/NetMQ/Core/Mechanisms/Mechanism.cs
@@ -350,12 +350,10 @@ namespace NetMQ.Core.Mechanisms
         
         protected bool IsCommand(string command, ref Msg msg)
         {
-            if (msg.Size >= command.Length + 1)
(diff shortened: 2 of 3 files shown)


drewnoakes:
The length check looks good, though adding a unit test for the `IsCommand` method seems like a good idea.

I'm less sure about the exception handling change. With this, failures go unreported and potentially undetected. Also, changing from `catch (TerminatingException)` to catching all exceptions also feels like a bad idea, at least by default. What is the actual scenario you're concerned about here?

Ideally this would be two PRs so we could discuss each change separately. They seem unrelated.

mnmr:
> The length check looks good, though adding a unit test for the `IsCommand` method seems like a good idea.

I have created some unit tests to validate the changes made. I started out by making tests to certify the old behavior, where NetMQ would throw an exception if the length byte in a command message was invalid. I then proceeded to verify that the new code works also in this scenario, without throwing an exception. I've included the test case proving the old behavior (uncommented) to make it easier for you to review.
 
> I'm less sure about the exception handling change. With this, failures go unreported and potentially undetected. Also, changing from `catch (TerminatingException)` to catching all exceptions also feels like a bad idea, at least by default. What is the actual scenario you're concerned about here?

Any uncaught exception in the Proactor.Loop method will cause a process exit, even if it caused by something benign like being unable to correctly parse incoming data (like in this scenario with the IsCommand method). Unless the failure is truly catastrophic, it is very unlikely that this behavior is desirable. And when things truly are catastrophic, chances are that things will fail elsewhere too, making it less important for this code to tear down the process.

The solution I'm working on makes use of long-lived services and having them abruptly exit because they receive something unepected on the wire is quite problematic. This change was meant to protect against any other undiscovered bugs that might trigger an exception.
 
> Ideally this would be two PRs so we could discuss each change separately. They seem unrelated.

I can split them up if you think it's better, but we are already using a custom-built NetMQ with the changes applied in production, so for me there is no immediate rush to get a new release out.

Irrespective of whether we split the PR or not, here are some possible paths forward to consider:

1. Revert the exception change and hope there isn't anything else that could potentially result in an exception.
2. Keep the change as-is, despite the mentioned flaws of catching everything.
3. Keep the change but add error reporting to surface any errors happening. Tthis would like imply referencing Microsoft.Extensions.Logging.Abstractions and passing an ILogger/ILogFactory on the ctor of various classes throughout NetMQ.
4. Keep the change but add a pluggable mechanism where users can inject or register an error handler (i.e. a method that consumes an Exception and reports back whether it has been handled or not; unhandled exceptions are then thrown). This allows us to keep the old behavior while still making exceptions catch-able for users that choose to provide a custom handler.

I think option 3 is the least attractive of these due to the number of files that likely have to be modified.

Option 4 is much less invasive, as this can be achieved with a static RegisterExceptionHandler on the Proactor, but it would still introduce new logic and thus potentially bugs. I'm not sure if there are other places in the code where a similar hook would be beneficial, but it seems like a thing you'd want to register once and have it apply everywhere relevant. Given my limited knowledge of the internals of NetMQ, going down this route would likely require someone to help me identify those places. This option likely also requires new documentation, so that users can find information on how it works and how to use it.

I'm biased towards option 2, obviously, with 1 being my second preference (mostly because options 3 and 4 seem like more work than justified for this), but let me know what you think and I will proceed accordingly.

drewnoakes:
Thanks very much for adding tests.

Crashing due to a bug in NetMQ is expected. The proper course of action is to fix such bugs. Swallowing exceptions makes them harder to discover and fix, and can lead to incorrect program behaviour. For critical systems it's better to crash than continue in an undefined state.

Catching all exceptions can break all kinds of scenarios and isn't something we should be doing.

I did consider some designs to surface these exceptions but because `Proactor` is an internal class it's harder to plumb these through to public API.

mnmr:
> Thanks very much for adding tests.

It's me thanking you for helping to get the fix published.
 
> Crashing due to a bug in NetMQ is expected. The proper course of action is to fix such bugs. Swallowing exceptions makes them harder to discover and fix, and can lead to incorrect program behaviour. For critical systems it's better to crash than continue in an undefined state.

I have reverted the commit that modified Proactor.

> I did consider some designs to surface these exceptions but because `Proactor` is an internal class it's harder to plumb these through to public API.

It could be as simple as a public static NetMQExceptionHandler class, where people can register the error handler and anything inside NetMQ that needs it can easily grab it. Though there may be other concerns I haven't considered. Maybe not important now that we're opting for option 1 :)

drewnoakes:
This looks great and I'll get it merged in.

If you wanted to create a new PR that explored having an opt-in way to handle exceptions that would otherwise crash the process, that would be great. As you mention, there could be a new static class. Or perhaps a static event on `NetMQConfig`. It'd have an event args with the exception, and a bool property to control whether to consider the exception handled or not (similar to `CancelEventArgs`).

src/NetMQ.Tests/MechanismTests.cs
```diff
@@ -0,0 +1,55 @@
...
+            msg.SetFlags(MsgFlags.Command);
+            msg.Put((byte)length);
+            msg.Put(Encoding.ASCII, data, 1);
+            return msg;
+        }
+
+        [Fact]
+        public void IsCommandShouldReturnTrueForValidCommand()
+        {
+            var mechanism = new NullMechanism(null, null);
```
drewnoakes, line 28:
```suggestion
            var mechanism = new NullMechanism(null!, null!);
```

src/NetMQ.Tests/MechanismTests.cs
```diff
@@ -0,0 +1,55 @@
...
+            var msg = CreateMsg("READY", 0);
+            Assert.True(mechanism.IsCommand("READY", ref msg));
+        }
+
+        [Fact]
+        public void IsCommandShouldReturnFalseForInvalidCommand()
+        {
+            var mechanism = new NullMechanism(null, null);
```
drewnoakes, line 36:
```suggestion
            var mechanism = new NullMechanism(null!, null!);
```

mnmr:
> If you wanted to create a new PR that explored having an opt-in way to handle exceptions that would otherwise crash the process, that would be great. As you mention, there could be a new static class. Or perhaps a static event on `NetMQConfig`. It'd have an event args with the exception, and a bool property to control whether to consider the exception handled or not (similar to `CancelEventArgs`).

Using NetMQConfig would make sense for this. That said, I think we need to run into another crash-causing bug before I'd go any further with this (we've been using NetMQ for years already, and it's generally quite stable so don't expect much benefit from this work).

drewnoakes:
Released in https://www.nuget.org/packages/NetMQ/4.0.1.11

Thanks again @mnmr!

Issue #1033, opened 2022-09-30
ZetMQ and Crestron Processors


JaronrH:
This is a VERY niche case but I still wanted to throw it out there!

I'm using ZeroMQ (NetMq 4.0.1.10) with a Crestron v4 Processor (they are used in home automation systems).  The processor uses Mono 6.12.0.107 for their .Net runtime with their latest firmware.  **Long story short: the native Opcode calls in NetMQ.Core.Utils fail causing the entire application to crash!** 

Thankfully, the fix was easy: Just remove the call to it in NetMQ.Core.Utils.Clock (much like your preprocessor command that singles out NETSTANDARD1_6) and everything worked as expected!

From this experience, I wanted to highlight 3 things for your consideration:
1) The error handling in NetMQ.Core.Utils.OpCode could use some improvement (there is none right now!).
2)  It could be helpful to allow NetMQ.Core.Utils.Clock.NowMs() to bypass NetMQ.Core.Utils.Clock.Rdtsc as an option (then I can bypass and avoid having a custom-built library!).
3)  The library needs some work to make it run in VS2022 (Include JetBrains.Annotations, Set the LangVersion to 8+, Fix Assumes.NotNull() calls in TcpListener.SetAddress() and V2Encoder.SizeReady(), etc.).  Nothing major but could use some attention!

Issue #714, opened 2017-11-28
netmq connection crash on NVidia TX1-TX2 systems


wavelet55:
### NVidia TX1 or TX2 computer

    NetMQ Version:   4.0.0.1 and 3.3.3.4 
    Operating System: ubuntu 16.04 LTS on ARMv8 Processor
    .NET Version:     4.5

### Expected behaviour
_cmdResponseSocket = new RequestSocket();
connStr = "tcp://192.168.1.10:5555"
_cmdResponseSocket.Connect(connStr);

Expect a connection... works on Windows and Ubuntu X86 machines.

### Actual behaviour
Causes a crash to the program on an NVidia TX1 & TX2 computer.
I trouble shot NetMQ and found issue in:
namespace NetMQ.Core.Utils Clock.cs 
The constructor:
        static Clock()
        {
#if !NETSTANDARD1_3 && !UAP
            try
            {
                if (Environment.OSVersion.Platform == PlatformID.Win32NT ||
                    Environment.OSVersion.Platform == PlatformID.Unix ||
                    Environment.OSVersion.Platform == (PlatformID)128)
                {
                    s_rdtscSupported = Opcode.Open();
                }
                else
                {
                    s_rdtscSupported = false;
                }
            }
            catch (Exception)
            {
                s_rdtscSupported = false;
            }
#endif
        }
works ok... s_rdtscSupported = Opcode.Open(); works and sets rdtscSupported = true.  There is a Opcode.Rdtsc  method attached.   The crash occurs when calling public static long Rdtsc().  The call to the Opcode.Rdtsc() causes the crash.
I assume the version of Linux on the NVidia TX1-TX2 systems has an issue.

I got around the problem by changing public static long Rdtsc() to:

public static long Rdtsc()
{
     return NowUs();
}

Issue #199, opened 2015-01-13
OpCode.Open() crash in Unity @ OSX


barzto:
Hi guys,

we're using NetMQ on Unity (primarily targeting iOS, but also running it on OSX desktop for testing/developing purposes, as it has shorter iteration time). We've encounter Unity crash after calling  (Dealer)Socket.Connect() when running on OSX (in Unity Editor/Desktop Player).

We've tracked the issue to `OpCode.Open()` method, namely to place where `mmap()` is called.
 
```C#
codeBuffer = (IntPtr)mmap.Invoke(null, new object[] { IntPtr.Zero, 
          size, mmapProtsParam, mmapFlagsParam, -1, 0 });
```
It fails for some unknown reason, the `codeBuffer` gets value = -1 (I guess this is the value of MMAP_FAILED). As the return value of `mmap` is not tested anywhere, it gets passed as destination address here:

```C#
Marshal.Copy(rdtscCode, 0, codeBuffer, rdtscCode.Length);
```
Which in turn crashes Unity Player.

I've patched our copy with following (being placed prior calling `Marshal.Copy()`), which basically turns off the RDSTC support.

```C#
                codeBuffer = (IntPtr)mmap.Invoke(null, new object[] { IntPtr.Zero, 
          size, mmapProtsParam, mmapFlagsParam, -1, 0 });

                if (codeBuffer == IntPtr.Zero || codeBuffer == (IntPtr)(-1))
                {
                    throw new InvalidOperationException("Mmap failed");
                }
```

It's not the best solution, but it do its job for us.

Seeing your assembler black magic there (one part of me really likes it, the other is pretty scared :) ), it brings one question to me: How this works on non x86 processors (like on ARM powering iOS devices). I guess it fails at some point and an exception is thrown, which disables RDTSC same way as my "fix", but I just one to bring this question up, so someone with better knowledge of NetMQ internals can properly answer it. 

Anyway great job with NetMQ, we're big fans!
Cheers,
-Jan

somdoron:
@barzto I'm not big fun of it myself, it's a port from the original project, I will look into removing it soon.

Can you provide a pull request with the patch you have made?

barzto:
Here we go: https://github.com/zeromq/netmq/pull/202

somdoron:
thanks

PR #1055, opened 2023-04-21
Fix fatal SEHException on Arm calling RDTSC

jonorossi:
Arm does not have the x86 Timestamp Counter register or read timestamp counter instruction. Add an additional architecture check to prevent the code path when running on Arm. The code checks for `X86`/`X64` as they are the only Intel-based architectures in the enum, while the enum now contains 4 Arm-based members (only in recent .NET) along with WebAssembly and more.

Running .NET on Windows on Arm, NetMQ throws a fatal exception crashing the application just starting up a poller:

```
Fatal error. System.Runtime.InteropServices.SEHException (0x80004005): External component has thrown an exception.
   at NetMQ.Core.Utils.Clock.Rdtsc()
   at NetMQ.Core.SocketBase.ProcessCommands(Int32, Boolean, System.Threading.CancellationToken)
   at NetMQ.Core.SocketBase.Bind(System.String)
   at NetMQ.NetMQSocket.Bind(System.String)
   at NetMQ.Sockets.PairSocket.CreateSocketPair(NetMQ.Sockets.PairSocket ByRef, NetMQ.Sockets.PairSocket ByRef)
   at NetMQ.Core.Utils.StopSignaler..ctor()
   at NetMQ.NetMQPoller..ctor()
   [...]
```

This defect has been reported in #714 (Arm), #199 (Arm), #714 (Arm) and #1033 (FPGA). There was one fix made in #227 that added the `IsARMArchitecture`, however that only works on Mono.

The change will fix the defect for .NET Standard 2.0/2.1 targets (which are used for .NET Core and 5+ runtimes) as they both define `NETSTANDARD1_1_OR_GREATER`. It won't fix the bug for .NET Framework 4.5/4.7 targets as the `RuntimeInformation` class is only available on .NET Framework 4.7.1, however Windows on Arm appears to emulate the instruction for .NET Framework and the code is set up ready if this project's TFMs are changed in the future.

### DIFF
--- a/src/NetMQ/Core/Utils/OpCode.cs
+++ b/src/NetMQ/Core/Utils/OpCode.cs
@@ -12,13 +12,21 @@ namespace NetMQ.Core.Utils
 
         public static bool Open()
         {
+#if NETSTANDARD1_1_OR_GREATER || NET471_OR_GREATER
+            if (RuntimeInformation.ProcessArchitecture != Architecture.X86 &&
+                RuntimeInformation.ProcessArchitecture != Architecture.X64)
+            {
+                return false; // RDTSC instruction not supported
+            }
+#endif
+
             var p = (int)Environment.OSVersion.Platform;
 
             byte[] rdtscCode = IntPtr.Size == 4 ? RDTSC_32 : RDTSC_64;
 
             s_size = (ulong)(rdtscCode.Length);
 
-            if ((p == 4) || (p == 128))
+            if ((p == 4) || (p == 128)) // Unix || Mono on Unix
             {
                 // Unix
                 if (IsARMArchitecture()) return false;


drewnoakes:
Thanks, this looks great.

Unfortunately, it looks like there's no way currently to run CI on ARM64.

drewnoakes:
Released in https://www.nuget.org/packages/NetMQ/4.0.1.12

jonorossi:
> Unfortunately, it looks like there's no way currently to run CI on ARM64.

I've pulled 4.0.1.12 and it runs. Cheers for making the release 👏.

drewnoakes:
Thanks for the high quality PRs and confirming the fix!

PR #1056, opened 2023-04-21
Remove .NET Standard 1.6 conditional code

jonorossi:
.NET Standard 1.6 support was removed in NetMQ 4.0.1.

* Remove unused conditional code for .NET Standard 1.6
* Keep `RuntimeInformation.IsOSPlatform` usage that only ran on .NET Standard 1.6 target


drewnoakes:
Thank you for this tidy up.

PR #1058, opened 2023-05-06
add net471 target

vashek:
This is to take advantage of https://github.com/zeromq/netmq/pull/1055

### DIFF
--- a/src/NetMQ/NetMQ.csproj
+++ b/src/NetMQ/NetMQ.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <Description>A 100% native C# port of the lightweight high performance messaging library ZeroMQ</Description>
     <Version>4.0.0.0</Version>
-    <TargetFrameworks>net45;net47;netstandard2.0;netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>net45;net47;net471;netstandard2.0;netstandard2.1</TargetFrameworks>
     <DebugType>portable</DebugType>
     <AssemblyOriginatorKeyFile>./NetMQ.snk</AssemblyOriginatorKeyFile>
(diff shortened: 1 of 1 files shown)


drewnoakes:
This looks good in theory, but I wonder if we can just bump the `net47` target to `net471` instead.

The downside is that anyone targeting `net47` would then drop to `net45`.

The upside is we have a simpler matrix of configurations to build/ship/test.

Does anyone have any thoughts around this?

vashek:
TBH I don't have an opinion either way, I don't consider myself any sort of expert on .NET and my experience with it is somewhat limited. I just find that taking advantage of the fix from #1055 - which happens to be essential for me and came at just the right moment, so huge thanks to @jonorossi - requires rebuilding NetMQ when the app targets .Net Framework higher than 4.7, so it would be quite convenient to have 4.7.1 (or I guess 4.8, in my case) included in the nuget package.

My backstory is that there's a proprietary app that uses NetMQ 3.3.3.4 and targets .Net Framework 4.8, and was made with little consideration for anything other than Windows on Intel, and I'm exploring making it run on a Raspberry using Mono. As is, it exhibits a native crash due to rdtsc. I found #1055 and so I tried a drop-in replacement of the NetMQ.dll with the 4.0.1.12 version, not really expecting to run at all (without rebuilding the app) - it ran, but with the same crash. So I rebuilt NetMQ for 4.7.1, replaced the DLL and voilà, it seems to run (or at least doesn't obviously crash, so far).

vashek:
Marked as ready to maybe grab more attention. ;)
Regarding the failed test I wonder if that's just a spurious error, something timed out in netcoreapp3.1, specifically `NetMQ.Tests.ClientServer.AsyncEnumerableCanceled`.

PR #1069, opened 2023-07-23
Implemented AsSpan on NetMQFrame

dxdjgl:


src/NetMQ/NetMQFrame.cs
```diff
@@ -337,5 +337,14 @@ public byte[] ToByteArray(bool copy = false)
...
             return byteArray;
         }
+
+        /// <summary>
+        /// Return an readonly span of bytes that carries the content of this NetMQFrames Buffer.
+        /// </summary>
+        /// <returns>the Buffer as a readonly span</returns>
+        public ReadOnlySpan<byte> ToReadOnlySpan()
+        {
+            return new ReadOnlySpan<byte>(Buffer);
```
drewnoakes, line 347: This should take the message length into account. The buffer could be longer than the message.

src/NetMQ/NetMQFrame.cs
```diff
@@ -337,5 +337,14 @@ public byte[] ToByteArray(bool copy = false)
 
             return byteArray;
         }
+
+        /// <summary>
+        /// Return an readonly span of bytes that carries the content of this NetMQFrames Buffer.
+        /// </summary>
+        /// <returns>the Buffer as a readonly span</returns>
+        public ReadOnlySpan<byte> ToReadOnlySpan()
```
drewnoakes, line 345: What do you think about calling this AsSpan, to match similar methods names in the BCL, such as on strings?

drewnoakes (commented):
Great idea, thanks.

dxdjgl:
You are absolutely right, review comments has been fixed.

dxdjgl:
@drewnoakes would you take another look at the code and let me know if other changes are needed.

drewnoakes (approved):
Apologies for the delay here. Looks great to me.

drewnoakes:
Available in https://github.com/zeromq/netmq/releases/tag/4.0.1.13

dxdjgl:
Thanks Drew for your help

PR #1072, opened 2023-08-31
Added a static Boolean on the SocketOptions class to avoid using the …

### LINKED ISSUE #1071: Antivirus dynamic shell code execution
### Environment

    NetMQ Version:    4.0.1.13
    Operating System: Windows 10
    .NET Version:     .Net 6

### Expected behavior
I would like to be able to have an option to force the NetMQ.Core.Utils.Clock's s_rdtscSupported to false, and therefore always fall back to using Stopwatch.GetTimestamp. As an idea add this Boolean:
```cs
public class SocketOptions
{
        /// <summary>
        /// If set, the time stamp counter is not read directly through opcode injection, rather
        /// <see cref="System.Diagnostics.Stopwatch.GetTimestamp"/> is used.
        /// </summary>
        public static bool DoNotUseRDTSC;
```
Then in Opcode.Open
```cs
        public static bool Open()
        {
            if (SocketOptions.DoNotUseRDTSC)
                return false;
```

### Actual behavior
 The method of injecting opcode into memory and invoking it to read the time stamp counter is being detected as dynamic shellcode by some anti-malware such as Sophos. When detected, the process executing the code is terminated. The detection as malware occurs when the Opcode.Open method is called and the s_codeBuffer is allocated with EXECUTE_READWRITE. 

Bind a publish socket to a tcp URL when running on a system protected by Sophos. Unfortunately, my companies corporate IT has deployed this feature everywhere and I can't argue it is able to detect actual malware scenarios. But I have a large engineering community where ZeroMQ is going to become more prevalent.

rickshaw5724:
…opcode injection technique for reading the time stamp counter.

Fixes #1071

### DIFF
--- a/src/NetMQ/SocketOptions.cs
+++ b/src/NetMQ/SocketOptions.cs
@@ -13,6 +13,16 @@ namespace NetMQ
     public class SocketOptions
     {
         /// <summary>
+        /// If set, the time stamp counter is not read directly through opcode injection,
+        /// rather <see cref="System.Diagnostics.Stopwatch.GetTimestamp"/> is used.
+        /// When false, the time stamp counter is read by allocating a few bytes on the heap with
+        /// read/write/execute privilege. OpCode is copied to this allocated memory and invoked to read
+        /// the time stamp counter, (which is a register available on most modern CPUs). While this is
+        /// an accurate way to read the time stamp counter, because it injects code onto the heap, this
+        /// can be detected as a malware technique by some anti-virus defenders.
+        /// </summary>
+        public static bool DoNotUseRDTSC;
+        /// <summary>
         /// The NetMQSocket that this SocketOptions is referencing.
         /// </summary>
         private readonly NetMQSocket m_socket;
--- a/src/NetMQ/Core/Utils/OpCode.cs
+++ b/src/NetMQ/Core/Utils/OpCode.cs
@@ -11,6 +11,8 @@ namespace NetMQ.Core.Utils
 
         public static bool Open()
         {
+            if (SocketOptions.DoNotUseRDTSC)
+                return false;
 #if NETSTANDARD1_1_OR_GREATER || NET471_OR_GREATER
             if (RuntimeInformation.ProcessArchitecture != Architecture.X86 &&
                 RuntimeInformation.ProcessArchitecture != Architecture.X64)


src/NetMQ/SocketOptions.cs
```diff
@@ -12,6 +12,16 @@ namespace NetMQ
...
+        /// <summary>
+        /// If set, the time stamp counter is not read directly through opcode injection,
+        /// rather <see cref="System.Diagnostics.Stopwatch.GetTimestamp"/> is used.
+        /// When false, the time stamp counter is read by allocating a few bytes on the heap with
+        /// read/write/execute privilege. OpCode is copied to this allocated memory and invoked to read
+        /// the time stamp counter, (which is a register available on most modern CPUs). While this is
+        /// an accurate way to read the time stamp counter, because it injects code onto the heap, this
+        /// can be detected as a malware technique by some anti-virus defenders.
+        /// </summary>
+        public static bool DoNotUseRDTSC;
```
drewnoakes, line 24: This doesn't really seem like a socket option, and we don't have any other static options here as far as I can tell.

Also the timing of setting this value is important, as the result of `Open` is cached in `Clock.s_rdtscSupported`.

Because of the timing issue, what about using something like an environment variable to control this. For example, if `NETQM_SUPPRESS_RDTSC` was present, it would disable any attempt to call `Opcode.Open()` and store a `false` in `s_rdtscSupported`. All that could happen in the static constructor of `Clock`.

rickshaw5724:
I agree SocketOptions isn't the right place. I'll change it to look for an
environment variable

debloke:
did this change make it to a nuget pkg?

drewnoakes:
The PR went stale, with unaddressed feedback. Feel free to submit a new PR that addresses it.

rickshaw5724:
I was brand new to git when I submitted the original issue #1071 It appears in the back and forth in getting the pull request correct, I created 2 pull requests, this one that went dead and the other one that actually fixed the issue and was merged back into the main branch on Sept 7, 2023. #1074 
As far as I can tell, there has not been a new NuGet package created since pull request #1074 was merged.

debloke:
the master indeed has the fix and seems to work. can we get the change pushed to nuget?

PR #1073, opened 2023-09-06
Bump versions and fix assertion

drewnoakes:


drewnoakes:
These changes have mostly (all?) been applied via other PRs.

PR #1074, opened 2023-09-06
Avoid reading time stamp counter

### LINKED ISSUE #1071: Antivirus dynamic shell code execution

rickshaw5724:
The implementations using an environment variable rather than a static flag on an unrelated class.

Fixes #1071

### DIFF
--- a/src/NetMQ/Core/Utils/OpCode.cs
+++ b/src/NetMQ/Core/Utils/OpCode.cs
@@ -11,6 +11,8 @@ namespace NetMQ.Core.Utils
 
         public static bool Open()
         {
+            if (SocketOptions.DoNotUseRDTSC)
+                return false;
 #if NETSTANDARD1_1_OR_GREATER || NET471_OR_GREATER
             if (RuntimeInformation.ProcessArchitecture != Architecture.X86 &&
                 RuntimeInformation.ProcessArchitecture != Architecture.X64)
--- a/src/NetMQ/Core/Utils/OpCode.cs
+++ b/src/NetMQ/Core/Utils/OpCode.cs
@@ -11,7 +11,7 @@ namespace NetMQ.Core.Utils
 
         public static bool Open()
         {
-            //Look for an environment variable: "NETQM_SUPPRESS_RDTSC" with any value
+            // Look for an environment variable: "NETQM_SUPPRESS_RDTSC" with any value.
             // The application can set this environment variable when this code is running in a system where
             // it is not desirable to read the processor's time stamp counter.
             // While this is supported in modern CPUs, the technique used for allocating executable memory, copying OP Code
--- a/src/NetMQ/Core/Utils/OpCode.cs
+++ b/src/NetMQ/Core/Utils/OpCode.cs
@@ -11,13 +11,14 @@ namespace NetMQ.Core.Utils
 
         public static bool Open()
         {
-            //Look for an environment variable: "NETQM_SUPPRESS_RDTSC" with a value of "TRUE"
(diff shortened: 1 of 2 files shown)


src/NetMQ/Core/Utils/OpCode.cs
```diff
@@ -11,6 +11,14 @@ internal static class Opcode
 
         public static bool Open()
         {
+            //Look for an environment variable: "NETQM_SUPPRESS_RDTSC" with a value of "TRUE"
+            // The application can set this environment variable when this code is running in a system where
+            // it is not desirable to read the processor's time stamp counter.
+            // While this is supported in modern CPUs, the technique used for allocating executable memory, copying OP Code
+            // for the read of the time stamp and invoking the OP Code can be detected as Malware by some anti-virus vendors.
```
drewnoakes, line 18: Please include a link to the original issue here so that we can more easily see the motivation for this feature in future:

```suggestion
            // for the read of the time stamp and invoking the OP Code can be detected as Malware by some anti-virus vendors.
            // https://github.com/zeromq/netmq/issues/1071
```

src/NetMQ/Core/Utils/OpCode.cs
```diff
@@ -11,6 +11,14 @@ internal static class Opcode
...
+            var val = Environment.GetEnvironmentVariable("NETQM_SUPPRESS_RDTSC");
+            if ("TRUE".Equals(val, StringComparison.OrdinalIgnoreCase))
+                return false;
```
drewnoakes, line 21: What do you think about supporting any value in this environment variable? It's not uncommon to want to set a variable like this to `1`. We could just check for any non-`null` value.

```suggestion
            if (Environment.GetEnvironmentVariable("NETQM_SUPPRESS_RDTSC") is not null)
                return false;
```

drewnoakes: The downside of this approach would be that setting the value to `false` would not do what you expect.

I'm happy with whatever you decide.

rickshaw5724:
I made the change to use any value set to the environment variable as suggested and added a link to the github issue.
I'm a bit new to github, so I hope this commits that to this pull request

drewnoakes (approved):
Looks great, thanks very much!

PR #1079, opened 2023-10-29
Update push-pull.md

apolloid:
the original example which uses a pull socket to send a frame will return an error since the pull socket does not have such function


drewnoakes (commented):
I'm not sure it's correct to remove the socket. I think instead it should be changed to a push socket.

See https://zguide.zeromq.org/docs/chapter1/#Divide-and-Conquer

apolloid:
Hi @drewnoakes, 
Yes, you are right about that and I have amended it.

drewnoakes:
Thanks!

PR #1095, opened 2024-06-07
Mitigating a FaultException in Mechanism.Encode.

### LINKED ISSUE #1094: Process crash due to an unhandled exception in Mechanism.Encode
### Environment

    NetMQ Version:     4.0.1.13
    Operating System: Windows Server 2019 Standard 64bit
    .NET Version:          .NET 6

We are using a component that creates a DealerSocket and use that component in a Windows service. The socket is listening on tcp address 0.0.0.0 and port 5556. It does so by issuing
`socket.TryReceiveMultipartMessage(ref clientMessage, 0)`
where clientMessage variable is of type of NetMQMessage and equals null.

### Expected behaviour
The system NetMQ socket that receives the data from the queue should not crash the process.

### Actual behaviour
Approximately once a day the Windows service process crashes due to an unhandled exception. The Exception with the message and the call stack is like the following:
```
Application: <wiped out by NDA>
CoreCLR Version: 6.0.1623.17311
.NET Version: 6.0.16
Exception Info: NetMQ.FaultException: Cannot close an uninitialised Msg.
   at NetMQ.Msg.Close() in /_/src/NetMQ/Msg.cs:line 453
   at NetMQ.Core.Mechanisms.CurveMechanismBase.Encode(Msg& msg) in /_/src/NetMQ/Core/Mechanisms/CurveMechanismBase.cs:line 66
   at NetMQ.Core.Transports.StreamEngine.PullAndEncode(Msg& msg) in /_/src/NetMQ/Core/Transports/StreamEngine.cs:line 1226
   at NetMQ.Core.Transports.StreamEngine.BeginSending() in /_/src/NetMQ/Core/Transports/StreamEngine.cs:line 445
   at NetMQ.Core.Transports.StreamEngine.Handle(Action action, SocketError socketError, Int32 bytesTransferred) in /_/src/NetMQ/Core/Transports/StreamEngine.cs:line 416
   at NetMQ.Core.Transports.StreamEngine.FeedAction(Action action, SocketError socketError, Int32 bytesTransferred) in /_/src/NetMQ/Core/Transports/StreamEngine.cs:line 333
   at NetMQ.Core.Transports.StreamEngine.OutCompleted(SocketError socketError, Int32 bytesTransferred) in /_/src/NetMQ/Core/Transports/StreamEngine.cs:line 1023
   at NetMQ.Core.IOObject.OutCompleted(SocketError socketError, Int32 bytesTransferred) in /_/src/NetMQ/Core/IOObject.cs:line 119
   at NetMQ.Core.Utils.Proactor.Loop() in /_/src/NetMQ/Core/Utils/Proactor.cs:line 130
```

This call stack corresponds to the code exactly because it is done in non-optimized version of NetMQ.dll.

The behavior is quite sporadic. The issue arises only approximately once a day. It is worth mentioning that in the same process we have another DealerSocket that sends messages to tcp port 127.0.0.1 and the address 5556.

While looking into the code it seems that on some circumstances StreamEngine recieves from the Pipe non-initialized incoming message. When it tries to pass it to Mechanism.Encode method from it's PullAndEncode() method, the non-initialized message throws an exception. I think on some circumstances the code in `NetMQ.Core.Utils.YQueue` can return a non-initialized message.

I am going to mitigate the issue by adding the additional check to `NetMQ.Core.SessionBase.PullMsg(ref Msg msg)`. Could you please consider merging this to the repo? I am going to provide a PR.

sonicpro:
On some circumstances Msg class throws NetMQ.FaultException in Mechanism.Encode method. The details are in https://github.com/zeromq/netmq/issues/1094

closes #1094

### DIFF
--- a/src/NetMQ/Core/SessionBase.cs
+++ b/src/NetMQ/Core/SessionBase.cs
@@ -217,7 +217,7 @@ namespace NetMQ.Core
         /// <returns>true if the Msg is successfully sent</returns>
         public virtual PullMsgResult PullMsg(ref Msg msg)
         {
-            if (m_pipe == null || !m_pipe.Read(ref msg))
+            if (m_pipe == null || !m_pipe.Read(ref msg) || !msg.IsInitialised)
                 return PullMsgResult.Empty;
             m_incompleteIn = msg.HasMore;
 


follesoe:
Any chance you could have a look at this PR, @drewnoakes? I know I am not entitled to nag or demand from open-source maintainers, but it would be great to get some updates on the NetMQ core. Right now, I am maintaining my own fork and building from source rather than using the NuGet package to get some of these fixes included.

drewnoakes:
While this does look like it'd suppress the exception, I'm not sure this is an actual fix. It might just allow the process to continue having skipped data or in some other invalid state, whereafter debugging the problem would probably be harder.

From the linked issue:

> I think on some circumstances the code in NetMQ.Core.Utils.YQueue can return a non-initialized message.

Indeed, looking at `YQueue` you can see it's not null-annotated, and there are a bunch of expectations around how the type is used. I wonder whether you'd be able to run a version of NetMQ with an implementation of `YQueue` that validates all nullness guarantees, to see if that's what's really going on. A process dump when the exception is thrown should provide insight into what state the application was in when the failure occurred.

https://stackoverflow.com/a/20238046/24874

A dump can be opened in Visual Studio to analyze the state of the process at the time of the crash. The instance of `YQueue`, `YPipe` and so on can be inspected to check if it's in a bad state.

I'm sympathetic to the problem here and want to find a solution that addresses the problem fully. It's just that the problem isn't well understood unfortunately.

sonicpro:
> While this does look like it'd suppress the exception, I'm not sure this is an actual fix. It might just allow the process to continue having skipped data or in some other invalid state, whereafter debugging the problem would probably be harder.
> 
> From the linked issue:
> 
> > I think on some circumstances the code in NetMQ.Core.Utils.YQueue can return a non-initialized message.
> 
> Indeed, looking at `YQueue` you can see it's not null-annotated, and there are a bunch of expectations around how the type is used. I wonder whether you'd be able to run a version of NetMQ with an implementation of `YQueue` that validates all nullness guarantees, to see if that's what's really going on. A process dump when the exception is thrown should provide insight into what state the application was in when the failure occurred.
> 
> https://stackoverflow.com/a/20238046/24874
> 
> A dump can be opened in Visual Studio to analyze the state of the process at the time of the crash. The instance of `YQueue`, `YPipe` and so on can be inspected to check if it's in a bad state.
> 
> I'm sympathetic to the problem here and want to find a solution that addresses the problem fully. It's just that the problem isn't well understood unfortunately.

Hello @drewnoakes I am going to collect the dump data on the program crash. But it is a windows service. Will the method with Windows Error Reporting by the link you posted https://stackoverflow.com/questions/20237201/best-way-to-have-crash-dumps-generated-when-processes-crash/20238046#20238046 work? I mean it is said in the post that Windows will collect the data only after you click "Close the program" in an error message that reports the program crash.

sonicpro:
Hi @drewnoakes. The issue stops reproducing. I failed to get the process dump on the program crash. I am closing this issue for now.

Issue #1109, opened 2024-12-19
Is the project dead?


LiorBanai:
- last pull is broken
- no update for the last 7 monts.
- no update of dependency vulnerability

![image](https://github.com/user-attachments/assets/4cd24c36-de28-4dca-970f-594e7f7f2c48)

LiorBanai:
sure, will do.

somdoron:
Thanks. We used to have a CI which doesn't seem to work anymore. We used
appveyor, we might want to move to github actions.

LiorBanai:
the ci is different on my own fork:
here it fails: https://github.com/zeromq/netmq/pull/1111
on my furk it passes: https://github.com/LiorBanai/netmq/pull/1

somdoron:
Can you create a new CI workflow to push a nuget version on tag or release?
I will add the nuget key to the project variables.

tmatthey:
@LiorBanai Thanks for the push! I've also an outstanding [PR 1093](https://github.com/zeromq/netmq/pull/1093) Since there was nothing happened I forked with my PR.

LiorBanai:
in my nugets workflow i just push the nuget always (when on master/main) with the skip duplicates flags. This allows the flexibility to update the version in the csproj/build.prop without needing a tag. 
@somdoron do yuo want a push only on tag or using the same approch?

LiorBanai:
@tmatthey  is that still relevant? maybe worth adding it before releasing new version..

somdoron:
We can push pre release on every commit and a final version on release/tag

LiorBanai:
Not sure why there is an issue:
 looking in the release.yml and prerelease.yml files, all the infrastructure is already there to push nugets so it should already work no?.

tmatthey:
Just one tiny comment on target frameworks; standard 2.0 is the last common target which both .Net Framework and .net can consume. Unfortunately, NetMQ has many dependencies where the standard 2.0 has vulnerabilities, which won't be fixed, I guess.

tmatthey:
And shouldn't the version be bumped up, minor? major? I would suggest major since some targets have be removed, breaking change.

LiorBanai:
Not sure a full CI replacement is needed as there does seem to be many changes in the library.

Can a new version be pushed with the current workflow and set up of new CI/CD be done after that?

Tanshaydar:
Anu update on this?
[...]

PR #1110, opened 2024-12-19
Remove end of life target frameworks

LiorBanai:
Remove the following:
- net45
- netsandard 2.0 (no longer recommended to use it)


LiorBanai:
part of #1109 
@somdoron  fyi.

this is technically breaking change as net45 is removed but that is very old version anyway

PR #1111, opened 2024-12-19
modernize repository

LiorBanai:
- [CY - YML] remove net3.1 and net 6 and update to use net9
- Update github actions to latest versions (checkout and setup-dotnet actions)


LiorBanai:
related to #1109

PR #1120, opened 2025-04-11
Tcp Keepalive for Linux (NET8.0)

chrbauer:
New Target .NET 8.0
Fixes setup of Tcp Keepalive under Linux

### DIFF
--- a/src/NetMQ.Tests/ClientServer.cs
+++ b/src/NetMQ.Tests/ClientServer.cs
@@ -82,7 +82,7 @@ namespace NetMQ.Tests
             await Assert.ThrowsAnyAsync<OperationCanceledException>(async () => await server.ReceiveStringAsync(source.Token));
         }
 
-#if NET8_0_OR_GREATER
+#if NETSTANDARD2_1_OR_GREATER
 
         [Fact(Timeout = 120)]
         public async void AsyncEnumerableCanceled()
--- a/src/NetMQ/Annotations.cs
+++ b/src/NetMQ/Annotations.cs
@@ -1,6 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE.md file in the project root for more information.
 
-#if !NETSTANDARD2_1 && !NET8_0_OR_GREATER
+#if !NET
 
 namespace System.Diagnostics.CodeAnalysis
 {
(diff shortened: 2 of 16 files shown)


AlexandreArpin:
I think adding targets to all currently supported net targets (so 8/9) would probably be a good thing, it should probably be done in another PR.

I maintain a fork since activity had slowed down on the repository and I wanted to "modernize" the repository a bit, I ended up completing the work of #1073 (and a bunch of smaller open PR) which I think is the right way about adding support to net target.

More precisely, a lot of the changes in this PR use the pragma `NET8_0_OR_GREATER` when they probably should use `NETSTANDARD2_1_OR_GREATER`.

It would be very nice to keep some of the momentum going from #1111

src/NetMQ/Core/Transports/Tcp/TcpConnector.cs
```diff
@@ -248,6 +248,13 @@ public void OutCompleted(SocketError socketError, int bytesTransferred)
                     // Set the TCP keep-alive option values to the underlying socket.
                     m_s.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, m_options.TcpKeepalive);
 
+                  
+#if NET8_0_OR_GREATER
```
AlexandreArpin, line 252: Last comment in the same "general" idea, these options became available in net5/netcore3.1, might be worth to do a `#if NET` here?

AlexandreArpin:
> Modernizing the project is a great idea. Even net6 had reached eol. The NETSTANDARD2_1_OR_GREATER could be indeed used on many places, but but not everywhere. The Pull request solves the core problem only for net8.
> 
> I would like to have this merged. Unsure how to proceed.

I'm in favor of adding net8/9 targets in additional to .net standard, although it should probably be done in its own PR and when the targets are added, then rebase this PR to only add the patch for the keep alive in the socket classes?

chrbauer:
I will prepare an update to set also TcpKeepaliveCnt...

chrbauer:
> I'm in favor of adding net8/9 targets in additional to .net standard, although it should probably be done in its own PR and when the targets are added, then rebase this PR to only add the patch for the keep alive in the socket classes?
I would be happy to do the rebase, if such a branch exists.

chrbauer:
> I will prepare an update to set also TcpKeepaliveCnt...

This was done last week

PR #1125, opened 2025-05-27
Draft: Support for disconnect messages socket option

chrbauer:


chrbauer:
I created a new pull request rebased against master

PR #1129, opened 2025-06-14
Review conditional code

drewnoakes:
Follows #1128


src/NetMQ.Tests/ClientServer.cs
```diff
@@ -82,10 +82,9 @@ public async Task AsyncWithCancellationToken()
             await Assert.ThrowsAnyAsync<OperationCanceledException>(async () => await server.ReceiveStringAsync(source.Token));
         }
 
-#if NETCOREAPP3_1
-
+#if NETSTANDARD || NET
```
AlexandreArpin, line 85: I still get tripped up by pragma defines but this pragma only enabled the test on net5+ (dotnet test 4.7.2 does not pick it up according to the CI logs)

I imagine this pragma was part of when certain targets didn't support IAsyncEnumerable? With netstandard2.0, net4.7.2 and net8+ supporting it, I imagine this pragma can removed let the tests run always?

AlexandreArpin (approved):
Super happy to see some activity on NetMQ

I double checked all the changes, lgtm!

src/NetMQ/Msg.cs
```diff
@@ -119,25 +116,25 @@ public struct Msg
         /// <summary>
         /// Returns true if msg is empty
         /// </summary>
-        public bool IsEmpty => m_data == null || Size == 0;
+        public readonly bool IsEmpty => m_data == null || Size == 0;
```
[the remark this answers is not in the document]
drewnoakes: Roslyn suggests adding `readonly` here, and it can theoretically help in some cases, depending on how the struct is used. It's not a bad thing to add IMO.

src/NetMQ.Tests/ClientServer.cs
drewnoakes: @chrbauer want to file that as a PR? Seems like a good change.

PR #1028, opened 2022-07-28
Update various documentation dead links and typos

AlexandreArpin:
I encountered a couple of dead links and typos in the documentation. I removed the 404 links and empty documentation pages.

There's a new version of mkdocs that added a breaking change to the `mkdocs.yml` file (pages becomes nav)

I think the readthedocs hasn't been refreshed/redeployed in a while because there's still an issue with the missing end code block on https://netmq.readthedocs.io/en/latest/message/ that was fixed in #1012

PR #1130, opened 2025-06-16
Update mkdocs

AlexandreArpin:
Revived #1028 - Trying to give some love to this project as I use it a lot

I'd like to get the documentation back to up and running - I ran mkdocs locally, can confirm it was broken but now it "Works On My Machine"™️after fixing the mkdocs.yaml file.

It's unclear if there's something more to be done for readthedocs to pick up the changes, I'm not 100% familiar with how it's hooked up under the hood but the mkdocs part is all fixed!

With a bit of luck, this is all that's needed and merging this updates readthedocs

### DIFF
--- a/docs/async-await.md
+++ b/docs/async-await.md
@@ -52,7 +52,7 @@ static void Main(string[] args)
 ```
 
 NetMQRuntime is a wrapper over NetMQPoller, when calling an async function the socket is automatically added to the internal poller.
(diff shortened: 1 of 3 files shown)


drewnoakes (approved):
Thanks for bringing this back!

drewnoakes:
Filed #1133 to get the service working again.

PR #1145, opened 2026-04-16
Upgrade net8.0 -> net10.0

chrbauer:
I think there is no reason to support net8.0 anymore, since the upgrade is trivial and the support is running out. There are still older NetMQ Packages for net8.0 available.

### DIFF
--- a/src/NetMQ.Tests/NetMQ.Tests.csproj
+++ b/src/NetMQ.Tests/NetMQ.Tests.csproj
@@ -7,7 +7,7 @@
     <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
     <IsTestProject>true</IsTestProject>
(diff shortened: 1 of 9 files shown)


chrbauer:
There is a security problem with System.Security.Cryptography.Xml. Overwrote the version (transitive deps). This is actually my main reason why i created the Pull request.

src/NetMQ.Tests/NetMQ.Tests.csproj
```diff
@@ -54,8 +54,8 @@
   </ItemGroup>
   
   <ItemGroup>
-    <PackageReference Update="JetBrains.Annotations" Version="2024.3.0" />
-    <PackageReference Update="Microsoft.SourceLink.GitHub" Version="8.0.0" />
+    <PackageReference Update="JetBrains.Annotations" Version="2025.2.4" />
+    <PackageReference Update="Microsoft.SourceLink.GitHub" Version="10.0.202" />
```
drewnoakes, line 58: The .NET SDK bundles this, since .NET 8.0. Better to just remove it so it doesn't get out of date.

PR #1157, opened 2026-05-01
Fix infinite blocking in TrySendMultipartMessage at low bandwidth

### LINKED ISSUE #1127: RouterSocket TrySendMultipartMessage SendMultipartMessage Hangs Infinitely With Low Network Speeds
I'm using a router dealer socket pair. I use NetLimiter to limit upload and download speeds to 1kbit/s SendMultipartMessage eventually hangs infinitely. I've also tried TrySendMultipartMessage with a timeout and it still hangs infinitely.

routerSocket.SendReady += (s, a) =>
{
    var netMqMessage = new NetMQMessage();
    netMqMessage.Append(socketIdentity);
    netMqMessage.AppendEmptyFrame();
    netMqMessage.Append(payload);
    a.Socket.SendMultipartMessage(netMqMessage);
};

VincentVbs:
Apply timeout to all frames in TrySendMultipartMessage, not just the first one. Previously, only the first frame used TrySendFrame with timeout, while subsequent frames used blocking SendMoreFrame/SendFrame methods. This caused infinite hangs when network bandwidth was extremely low (< 10 kbit/s).

The fix ensures all frames respect the timeout parameter, preventing infinite blocking and allowing graceful failure when send operations cannot complete.

Added three unit tests to validate the fix:
- TrySendMultipartMessage_AppliesTimeoutToAllFrames
- TrySendMultipartMessage_FailsGracefullyOnMiddleFrameTimeout
- TrySendMultipartMessage_FailsGracefullyOnLastFrameTimeout

Fixes #1127

### DIFF
--- a/src/NetMQ/OutgoingSocketExtensions.cs
+++ b/src/NetMQ/OutgoingSocketExtensions.cs
@@ -456,25 +456,17 @@ namespace NetMQ
         {
             if (message.FrameCount == 0)
                 throw new ArgumentException("message is empty", nameof(message));
-            else if (message.FrameCount == 1)
+
+            // Send all frames except the last with timeout and 'more' flag
+            for (int i = 0; i < message.FrameCount - 1; i++)
             {
-                return TrySendFrame(socket, timeout, message[0].Buffer, message[0].MessageSize);
-            }
-            else
-            {
-                bool sentSuccessfully = TrySendFrame(socket, timeout, message[0].Buffer, message[0].MessageSize, true);
+                bool sentSuccessfully = TrySendFrame(socket, timeout, message[i].Buffer, message[i].MessageSize, true);
                 if (!sentSuccessfully)
                     return false;
             }
 
-            for (int i = 1; i < message.FrameCount - 1; i++)
-            {
-                socket.SendMoreFrame(message[i].Buffer, message[i].MessageSize);
-            }
-
-            socket.SendFrame(message.Last.Buffer, message.Last.MessageSize);
-
-            return true;
+            // Send the last frame with timeout (no 'more' flag)
+            return TrySendFrame(socket, timeout, message.Last.Buffer, message.Last.MessageSize, false);
         }
 
         #endregion
--- a/src/NetMQ.Tests/OutgoingSocketExtensionsTests.cs
+++ b/src/NetMQ.Tests/OutgoingSocketExtensionsTests.cs
@@ -345,5 +345,110 @@ namespace NetMQ.Tests
 
 	        Assert.True(socket.TrySendFrame(TimeSpan.Zero, buffer, data.Length));
         }
+
+        /// <summary>
+        /// Test that TrySendMultipartMessage applies timeout to ALL frames, not just the first.
+        /// This validates the fix for issue #1127 where low bandwidth caused infinite blocking
+        /// because only the first frame respected the timeout parameter.
+        /// </summary>
+        [Fact]
+        public void TrySendMultipartMessage_AppliesTimeoutToAllFrames()
+        {
+            var timeout = TimeSpan.FromMilliseconds(500);
+            var frameCount = 0;
+
+            var socket = new MockOutgoingSocket((ref Msg msg, TimeSpan actualTimeout, bool more) =>
+            {
+                // CRITICAL: ALL frames must respect the timeout parameter
+                Assert.Equal(timeout, actualTimeout);
+                
+                frameCount++;
+                
+                // Verify 'more' flag is correct for each frame
+                if (frameCount < 3)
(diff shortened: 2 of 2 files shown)


src/NetMQ/OutgoingSocketExtensions.cs
```diff
@@ -456,25 +456,17 @@ public static bool TrySendMultipartMessage(this IOutgoingSocket socket, TimeSpan
...
-            {
-                return TrySendFrame(socket, timeout, message[0].Buffer, message[0].MessageSize);
-            }
-            else
+
+            // Send all frames except the last with timeout and 'more' flag
+            for (int i = 0; i < message.FrameCount - 1; i++)
             {
-                bool sentSuccessfully = TrySendFrame(socket, timeout, message[0].Buffer, message[0].MessageSize, true);
+                bool sentSuccessfully = TrySendFrame(socket, timeout, message[i].Buffer, message[i].MessageSize, true);
```
drewnoakes, line 463: This was in the old code too, but a named argument would be nice here.

```suggestion
                bool sentSuccessfully = TrySendFrame(socket, timeout, message[i].Buffer, message[i].MessageSize, more: true);
```

src/NetMQ/OutgoingSocketExtensions.cs
```diff
@@ -456,25 +456,17 @@ public static bool TrySendMultipartMessage(this IOutgoingSocket socket, TimeSpan
...
-            for (int i = 1; i < message.FrameCount - 1; i++)
-            {
-                socket.SendMoreFrame(message[i].Buffer, message[i].MessageSize);
-            }
-
-            socket.SendFrame(message.Last.Buffer, message.Last.MessageSize);
-
-            return true;
+            // Send the last frame with timeout (no 'more' flag)
+            return TrySendFrame(socket, timeout, message.Last.Buffer, message.Last.MessageSize, false);
```
drewnoakes, line 469:
```suggestion
            return TrySendFrame(socket, timeout, message.Last.Buffer, message.Last.MessageSize, more: false);
```

drewnoakes (approved):
Thanks, this looks great too.

drewnoakes:
Looks like there's a test failure. I like where this is going.

```
The active test run was aborted. Reason: Test host process crashed : Unhandled exception.
System.Net.Sockets.SocketException (22): Invalid argument
   at System.Net.Sockets.Socket.UpdateStatusAfterSocketErrorAndThrowException(SocketError error, Boolean disconnectOnFailure, String callerName)
   at System.Net.Sockets.Socket.UpdateStatusAfterSocketOptionErrorAndThrowException(SocketError error, String callerName)
   at System.Net.Sockets.Socket.SetSocketOption(SocketOptionLevel optionLevel, SocketOptionName optionName, Int32 optionValue, Boolean silent)
   at NetMQ.Core.Transports.Tcp.TcpConnector.OutCompleted(SocketError socketError, Int32 bytesTransferred) in /_/src/NetMQ/Core/Transports/Tcp/TcpConnector.cs:line 240
   at NetMQ.Core.Utils.Proactor.Loop() in /_/src/NetMQ/Core/Utils/Proactor.cs:line 130
```

drewnoakes:
@VincentVbs are you interested in continuing with this PR?

VincentVbs:
Thanks, @drewnoakes, for following up with me and for giving me the opportunity to contribute to this project.
Unfortunately, even though I spent extra time on the PR, I’m not able to do more today.
I did not anticipate running the unit tests on my machine, and the fix I initially proposed, which seemed trivial, turned out to be unfounded after reviewing the failing unit test.
In the end, I wasn’t able to reproduce the bug, but I have the feeling that the issue might no longer exist in the latest sources.
It would be useful to check exactly which version the user experienced the infinite loop with.
I did some tests using a Debian virtual machine and a tool to slow down communication, but I admit I wasn’t systematic enough and ended up getting exhausted.
I’m also not sure whether using a VM is the right approach to recommend.
Sorry for the noise this may have caused in the project.
Maybe I should have asked for guidance before working on the code.
How should I come back to zmq for a future contribution?
Thank you for taking the time to read this.

drewnoakes:
No problem at all @VincentVbs, it's great that you added to the project. Even incomplete PRs can help move things along. 

> Maybe I should have asked for guidance before working on the code.

Some projects ask contributors to do that, but generally I think the main reason to do that is to protect your own time against working on something that the project won't accept. Sometimes it's hard to know whether it's acceptable before the investigation is finished though. Don't worry about it.

> How should I come back to zmq for a future contribution?

Just work on anything that seems interesting to you. You can discuss beforehand if you like. We're happy to have any contribution here on NetMQ. Other projects in the zeromq org may have different requirements.

PR #1158, opened 2026-05-01
Fix ARM architecture detection for aarch64

### LINKED ISSUE #1116: Rdtsc crashing on ARM after Ubuntu 16bit to 32bit update
NetMQ.Core.Utils.Clock.Rdtsc is crashing after updating from Ubuntu 16.04 16bit version to Ubuntu 24.04 32bit version on a Raspberry PI 3b.
It is related to the checking of the processor architecture in IsARMArchitecture().

OpCode.cs:
```
 var machineValue = (string)utsname.GetField("machine").GetValue(currentValues);
            return machineValue.ToLower().Contains("arm");
```
In Ubuntu 16.04 16bit version the  machineValue was "armv71", now in Ubuntu 24.04 32bit version machineValue is "aarch64"

The simple fix would be:
```
 var machineValue = (string)utsname.GetField("machine").GetValue(currentValues);
            return machineValue.ToLower().Contains("aarch") || machineValue.ToLower().Contains("arm");
```

VincentVbs:
Add detection for "aarch" prefix to support modern 64-bit ARM architectures.

The IsARMArchitecture() method only checked for "arm" in the machine value, which failed to detect aarch64 (ARM 64-bit) architectures. This caused NetMQ to attempt using x86/x64 RDTSC instruction on ARM systems, resulting in crashes.

After Ubuntu upgrade from 16.04 (armv7l) to 24.04 (aarch64) on Raspberry Pi, the detection failed and NetMQ crashed with IllegalInstructionException.

Fix checks for both "arm" and "aarch" prefixes, supporting:
- Legacy 32-bit ARM: armv6l, armv7l, armhf (still works)
- Modern 64-bit ARM: aarch64, aarch32 (now works)

Tested with 12 unit tests covering all ARM variants and non-ARM architectures.

Fixes #1116

### DIFF
--- a/src/NetMQ/Core/Utils/OpCode.cs
+++ b/src/NetMQ/Core/Utils/OpCode.cs
@@ -95,7 +95,7 @@ namespace NetMQ.Core.Utils
 
             var currentValues = parameters[0];
             var machineValue = (string)(utsname!.GetField("machine")?.GetValue(currentValues) ?? "unknown");
-            return machineValue.ToLower().Contains("arm");
+            return machineValue.ToLower().Contains("arm") || machineValue.ToLower().Contains("aarch");
         }
 
         public static void Close()


src/NetMQ/Core/Utils/OpCode.cs
```diff
@@ -95,7 +95,7 @@ private static bool IsARMArchitecture()
 
             var currentValues = parameters[0];
             var machineValue = (string)(utsname!.GetField("machine")?.GetValue(currentValues) ?? "unknown");
-            return machineValue.ToLower().Contains("arm");
+            return machineValue.ToLower().Contains("arm") || machineValue.ToLower().Contains("aarch");
         }
```
drewnoakes, line 99: In modern .NET, we can use something like:

```csharp
bool isArmMachine = RuntimeInformation.OSArchitecture is Architecture.Arm or Architecture.Arm64;
```

We might need some conditional compilation for that, as we still target older .NET Framework versions.

drewnoakes (approved):
Looks great. One suggestion to make this more robust on modern .NET.

drewnoakes:
@VincentVbs are you interested in continuing with this PR?

VincentVbs:
Sorry, no

